PR #6, opened 2016-08-03
request and response

LoganWright:


http/request.md
```diff
@@ -0,0 +1,216 @@
...
+```swift
+public var method: Method
+public var uri: URI
+public var parameters: Node
+public var headers: [HeaderKey: String]
+public var body: Body
+public var data: Content
+```
+
+#### Method
```
tannernelson: `##` and `###` these have the best styles in the web UI.

Here is an example of #, ##, and ###
screen shot 2016-08-03 at 6 50 45 pm

http/request.md
```diff
@@ -0,0 +1,216 @@
...
+
+For example, say I receive a request to `http://vapor.codes?hello=world`.
+
+```swift
+let world = request.data["hello"].string
+```
+
+This same code will work if I receive a JSON request, for example:
+
+```
```
tannernelson: json

tannernelson:
Amazing

PR #36, opened 2016-09-18
create routing controller

VikinArtisan:
Will continue to improve

### DIFF
--- a/routing/controller
+++ b/routing/controller
@@ -47,7 +47,28 @@ This single route declaration creates multiple routes to handle a variety of act
 The generated controller will already have methods stubbed for each of these actions, including 
 notes informing you of the HTTP verbs and URIs they handle.
 
-| Verb          | URI           | Action|
-| ------------- |:-------------:| -----:|
+| Verb            | URI             |  Action       |
+| :-------------: | :-------------: | :-----------: |
+| GET             | test/index      | test.index    |
+| POST            | test/create     | test.create   |
+| GET             | test/show       | test.show     |
+| PUT             | test/replace    | test.replace  |
+| PATCH           | test/destroy    | test.destroy  |
+| DELETE          | test/destroy    | test.destroy  |
+| DELETE          | test/clear      | test.clear    |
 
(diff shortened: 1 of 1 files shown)


routing/controller
```diff
@@ -0,0 +1,53 @@
+---
+currentMenu: routing-controller
+---
+
+# Introduction
+
+Instead of defining all of your request handling logic as Closures in route files, you may wish to organize this behavior
+using Controller classes. Controllers can group related request handling logic into a single class. Controllers are stored
```
LoganWright: I think the newlines here and before `... in the Sources/App` section will format weird on some screens, let's just keep it one continuous line and let the browsers wrap it for us.

routing/controller
```diff
@@ -0,0 +1,53 @@
...
+```
+You can define a route to this controller action like so:
+
+```swift
+drop.get("getindex") {request in
+    return try FirstController().index(request: request)
+}
+```
+Now, when a request matches the specified route URI, the Index method on the FirstController 
+class will be executed. Of course, the route parameters will also be passed to the method.
```
LoganWright: I think it might be nice to demonstrate, or at the very least mention, that we could automate this behavior together by doing. `drop.get("getIndex", handler: FirstController().index)`

routing/controller
```diff
@@ -0,0 +1,53 @@
...
+Vapor resource routing assigns the typical "CRUD" routes to a controller with a single line of code. 
+
+```swift
+drop.resource("URI", Controller())
+```
+This single route declaration creates multiple routes to handle a variety of actions on the resource. 
+The generated controller will already have methods stubbed for each of these actions, including 
+notes informing you of the HTTP verbs and URIs they handle.
+
+| Verb          | URI           | Action|
```
LoganWright: Is this table unfinished, maybe we should pull it out.

VikinArtisan:
@LoganWright Would you please review!

LoganWright:
@VikinArtisan looks good, thank you much!

PR #40, opened 2016-09-22
update hello world

samisuteria:
I ran into this issue when using autocomplete for the `.get` method on `Droplet` in Xcode. Its an easy fix but I feel like it should be in the documentation. 

After: 
```swift
import Vapor
import HTTP

let drop = Droplet()

drop.get("hello") { (request: Request) -> ResponseRepresentable in
    return "Hello Swift"
}

drop.get { req in
    let lang = req.headers["Accept-Language"]?.string ?? "en"
    return try drop.view.make("welcome", [
    	"message": Node.string(drop.localization[lang, "welcome", "title"])
    ])
}

drop.resource("posts", PostController())

drop.run()
```

### DIFF
--- a/getting-started/hello-world.md
+++ b/getting-started/hello-world.md
@@ -61,11 +61,19 @@ drop.get("hello") { request in
 }
 ```
 
+Also add to the top of the file
+
+```swift
+import HTTP
+```
+
 This creates a new route on the `Droplet` that will match all `GET` requests to `/hello`.
 
 All route closures are passed an instance of [Request](../guide/request.html) that contains information such as the URI requested and data sent.
 
 This route simply returns a string, but anything that is [ResponseRepresentable](../guide/routing.html) can be returned. Learn more in the [Routing]({../guide/routing.html) section of the guide.
+
+Importing HTTP exposes the [ResponseRepresentable](../guide/routing.html) and [Request](../guide/request.html) classes
 
 ## Running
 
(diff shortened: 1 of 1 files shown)


tannernelson:
@samisuteria what about the autocomplete wasn't working before? You shouldn't need `(request: Request) -> ResponseRepresentable in` in the closure.

samisuteria:
When you let Xcode autocomplete it puts the types into the closure: 

screenshot 2016-09-21 21 57 18

screenshot 2016-09-21 21 57 49

which leads to the following errors unless you `import HTTP`

screenshot 2016-09-21 22 00 40

tannernelson:
Just do this:

```swift
drop.get("hello") { req in
    return "Hello Swift"
}
```

samisuteria:
Oh yea I'm aware that works. I was only suggesting that change for those that like to have explicit types in their closures :)

tannernelson:
Oh I see what you're saying. That seems more like an issue with Xcode making those ugly autocompletes. Having the additional metadata around the closure input variables makes the code harder to read. I'd rather put a note about how to remove that extraneous autocomplete data than have people keeping that in their source code.

tannernelson:
Something like:

```
Note: Xcode autocomplete may add extraneous type information to your closure's input arguments. This can be deleted to make the code cleaner.
```

tannernelson:
Thanks!

PR #55, opened 2016-10-11
Updated to correct directory structure Sources>Apps

symonty:
The documentaiton stated >App>main.swift
when I did the process i noted it was >Sources>App>main.swift

Confused me for a bit so I thought it was worth an update


LoganWright:
@symonty you're awesome, great catch!

PR #58, opened 2016-10-14
fix drop.get(_:, _: ) -> drop.get(_:, handler: )

iq3addLi:


LoganWright:
@iq3addLi thank you for adding this 👍

PR #59, opened 2016-10-14
Added SwiftyBeaver to provider list in docs

skreutzberger:


skreutzberger:
it is already in now so the PR can be closed.

PR #60, opened 2016-10-14
[BasicRouting] hyperlinks `Method` to 'response.md'

KevinVitale:
Quality-of-life change. 

> _Although, would've preferred to have been able to jump directly to `response.html#Method`. Not sure if Markdown can support that with relative links._


KevinVitale:
Whoa. I totally misunderstand this _simply_ line of documentation. A __Swift__ method; my bad ☹️

PR #67, opened 2016-10-30
Added information regarding sudo usage

reza-ryte-club:


reza-ryte-club:
Adding documentation regarding this issue:

[https://github.com/vapor/vapor/issues/675](url)

LoganWright:
@reza-ryte-club thank you for adding this. Very much appreciated!

amraboelela:
Better solution is to do ```sudo ln -s /opt/swift/usr/bin/swift /usr/bin/swift```

PR #69, opened 2016-11-01
Add VaporS3Signer Provider

JustinM1:


LoganWright:
@JustinM1 thanks for your support!

PR #108, opened 2017-01-11
Module Testing

mcdappdev:
Added documentation about how to break the app up into multiple modules for in-line Xcode testing, and just overall better practice.

### DIFF
--- a/couscous.yml
+++ b/couscous.yml
@@ -154,6 +154,9 @@ menu:
         testing:
           name: Testing
           items:
+              testing-modules:
+                  text: Modules
+                  relativeUrl: testing/modules.html
               testing-basic:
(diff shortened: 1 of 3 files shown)


mcdappdev:
I was also thinking about adding a new section about using MySQL with Travis/running tests on there. Any thoughts on that?

mcdappdev:
👍  Thanks! I'll make whatever English changes necessary.

BrettRToomey:
It's not really a big deal, it's just that we should pick one context to address the reader by and use it throughout the entire page :)

mcdappdev:
Agreed! I'd love to hear input from Tanner and Logan too to see if they have a preference over which tense.

PR #116, opened 2017-02-04
Download shell scripts through HTTPS (requires vapor.sh server config fix!)

tuxtina:
In several places in the docs, users are instructed to download a shell script and pipe it straight to `bash`:

    curl -sL check.vapor.sh | bash
    curl -sL toolbox.vapor.sh | bash
    curl -sL swift.vapor.sh/ubuntu | bash

While this is pretty and easy to remember, it also causes the initial connection to vapor.sh to be made insecurely through HTTP. A man-in-the-middle attack could intercept the connection and respond with a malicious shell script that would then be executed by bash.

Changing the commands to 

    curl -sL https://check.vapor.sh | bash
    curl -sL https://toolbox.vapor.sh | bash
    curl -sL https://swift.vapor.sh/ubuntu | bash

will cause the connecion to be made through HTTPS.

Unfortunately, only swift.vapor.sh seems to have a working SSL configuration at this time. The "check" and "toolbox" subdomains use the certificate for swift.vapor.sh and return a HTTP 500 error if connecting anyway. **This needs to be fixed before this pull request can be merged.**

### DIFF
--- a/getting-started/install-swift-3-macos.md
+++ b/getting-started/install-swift-3-macos.md
@@ -21,7 +21,7 @@ After Xcode 8 has been downloaded, you must open it to finish the installation. 
 Double check the installation was successful by running:
 
 ```sh
-curl -sL check.vapor.sh | bash
+curl -sL https://check.vapor.sh | bash
 ```
 
 ## Toolbox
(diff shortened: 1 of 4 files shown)


LoganWright:
@tuxtina thank you!

tuxtina:
@LoganWright It seems that the SSL config on vapor.sh for the check.vapor.sh and toolbox.vapor.sh subdomains does not yet have the correct config, so the HTTPS URLs don't work yet. See also https://github.com/vapor/toolbox/pull/134

PR #117, opened 2017-02-05
Fixed error with incorrect variable for refering to JSON Bytes

doronkatz:
Fixed text issue referring to incorrect variable for JSONBytes


LoganWright:
Thanks @doronkatz :)

PR #125, opened 2017-02-27
Added the node: attribute to the client POST x-www-form-urlencoded ex…

cyberdude:
Added the node: attribute to the client POST x-www-form-urlencoded example. It’s needed if you are going to use a variable as a value.  Ex: Body.data(Node(node: [“code”:code]


LoganWright:
@cyberdude thanks for the addition 👍

PR #214, opened 2017-08-25
Simplify Homebrew installation

vinkla:
Homebrew will automagically tap if one installs a binary with the full tap path. This pull request removes the step of tapping.


vinkla:
Closing due to inactivity.

PR #235, opened 2017-11-05
Redis Pipeline documentation

John-Connolly:


Joannis:
You should also add it to [the index of mkdocs](https://github.com/vapor/documentation/blob/vapor3/3.0/mkdocs.yml#L36), looks good 👍  thanks for all the help

John-Connolly:
Should be good to go now!

PR #240, opened 2017-11-16
Fix typo in leaf/leaf

albinekcom:


Joannis:
In what sense is this a typo? I might be overlooking it

albinekcom:
The result of `#()hashtags` is `#hashtags`, not `#Hashtags` (the first letter should not be uppercased).

Joannis:
Oh, got it! thanks 👍

PR #241, opened 2017-11-16
Enhancing documentation for Vapor Controller

martinpilch:
Reading through the documentation I was struggling with Controllers section trying to figure out how to test them.

I extended it with sample calls to demonstrate how to test them locally.

I also fixed error in declaration of `func sayHelloAlternate` to let it throw an error since `try` is called in it's body.

### DIFF
--- a/2.0/docs/vapor/controllers.md
+++ b/2.0/docs/vapor/controllers.md
@@ -35,7 +35,7 @@ let hc = HelloController()
 drop.get("hello", handler: hc.sayHello)
 ```
 
-Since the signature of our `sayHello` method matches the signature of the closure for the `drop.get` method, we can pass it directly. If you want to test it locally simply call `http://0.0.0.0:8080/hello?name=John`.
+Since the signature of our `sayHello` method matches the signature of the closure for the `drop.get` method, we can pass it directly. If you want to test it locally simply call `http://localhost:8080/hello?name=John`.
 
 ### Type Safe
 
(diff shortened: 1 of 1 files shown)


Joannis:
Thanks for the fix 👍 I'm wondering if `0.0.0.0` works on Linux. I usually browse to `localhost`

martinpilch:
@Joannis Good point, I change it to `localhost`

2.0/docs/vapor/controllers.md
```diff
@@ -35,7 +35,7 @@ let hc = HelloController()
 drop.get("hello", handler: hc.sayHello)
 ```
 
-Since the signature of our `sayHello` method matches the signature of the closure for the `drop.get` method, we can pass it directly.
+Since the signature of our `sayHello` method matches the signature of the closure for the `drop.get` method, we can pass it directly. If you want to test it locally simply call `http://localhost:8080/hello?name=John`.
```
Joannis: I just re-read these lines. Is "calling" a route the right term? I think "opening" the URL, or something similar might be more clear. We could also use markdown links with `[http://localhost:8080/hello?name=john](http://localhost:8080/hello?name=john)` so it's clickable

Feel free to change this or leave it be. I'll merge this PR tomorrow either way.

martinpilch: Makes sense 👍 Fixed

PR #249, opened 2017-12-03
Update install instructions.

Zhukn1:
With the latest version of brew on MacOS 10.13.1 I get:
`Error: No available formula with the name "vapor"`

`brew install vapor/tap/vapor` works fine

source: https://stackoverflow.com/a/43071459/573938

### DIFF
--- a/2.0/docs/getting-started/install-on-macos.md
+++ b/2.0/docs/getting-started/install-on-macos.md
@@ -50,7 +50,7 @@ brew update
 Now that you've added Vapor's tap, you can install Vapor's toolbox and dependencies.
 
 ```sh
-brew install vapor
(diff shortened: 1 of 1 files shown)


2.0/docs/getting-started/install-on-macos.md
```diff
@@ -50,7 +50,7 @@ brew update
 Now that you've added Vapor's tap, you can install Vapor's toolbox and dependencies.
 
 ```sh
-brew install vapor
+brew install vapor/tap/vapor
```
0xTim: This shouldn't be needed if you've followed the instructions above about adding the Vapor tap - though I agree the whole section could probably just be inlined

PR #253, opened 2017-12-13
Update supervisor.md

scauos:
I used supervisor to manage the process, in the config file, it should be 
```
[program:HelloVapor]
command=/home/HelloVapor/.build/debug/Run serve --env=production
directory=/home/HelloVapor
user=root
stdout_logfile=/var/log/supervisor/%(program_name)-stdout.log
stderr_logfile=/var/log/supervisor/%(program_name)-stderr.log
```
because when you use `command=/home/HelloVapor/.build/release/Run serve --env=production`
will failed with the error "could not find file in command=/home/HelloVapor/.build/release/Run"

### DIFF
--- a/2.0/docs/deploy/supervisor.md
+++ b/2.0/docs/deploy/supervisor.md
@@ -15,7 +15,7 @@ Each Vapor app on your server should have its own configuration file. For an exa
 
 ```sh
 [program:hello]
-command=/home/vapor/hello/.build/release/Run serve --env=production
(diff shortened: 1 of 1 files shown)


Joannis:
@scauos for production environment it's recommended to compile using `swift build -c release`. Release compiled binaries get put in the `.build/release/` and have been between 10 and 300x more performant in my own tests.

scauos:
Cool, Thanks

PR #254, opened 2017-12-21
Fix typo

twof:
RespondeEncodable -> ResponseEncodable


Joannis:
Thank you! I really wouldn't have found that

PR #256, opened 2017-12-21
First pass at Leaf documentation for 3.0.

twostraws:
This provides initial documentation for Leaf in Vapor 3.0. I've re-arranged it a little from the 2.0 documentation so that it provides examples of routing code up front so that readers have something to help them get started. I've also added specific sections for conditions, loops, and embedding, as they are common tasks.


3.0/mkdocs.yml
```diff
@@ -73,6 +73,10 @@ pages:
 #         - 'Package': 'databases/postgres/getting-started.md'
 #         - 'Setup': 'databases/postgres/setup.md'
 #         - 'Basics': 'databases/postgres/basics.md'
+- 'Leaf':
+    - 'Package': 'leaf/package.md'
```
Joannis: `Package` should be renamed to `Getting Started` (getting-started.md) to stay in tune with the rest.

twostraws: Right now it looks like other modules use "Package" for the visible name and getting-started.md for the filename, so I'll mimic that.

PR #260, opened 2017-12-22
Typo fix

twof:
`enum`s makes sense -> `enum`s make sense


Joannis:
What changed?

twof:
"However, for something like a supported data type, `enum`s makes sense because"
changed to
"However, for something like a supported data type, `enum`s **make** sense because"

PR #263, opened 2017-12-22
Fix build errors

revolter:


0xTim:
Again, the PR has a mix of Swift 3 and Swift 4 syntax, hopefully it will be consistent for Swift 5! But we probably need to provide both for Vapor 2

PR #265, opened 2017-12-22
Fix not being able to import AuthProvider

revolter:
Did anyone event test these docs?

### DIFF
--- a/2.0/docs/auth/package.md
+++ b/2.0/docs/auth/package.md
@@ -7,6 +7,8 @@ This section outlines how to import the Auth package both with or without a Vapo
 The easiest way to use Auth with Vapor is to include the Auth provider. 
 
 ```swift
+// swift-tools-version:4.0
+
 import PackageDescription
 
 let package = Package(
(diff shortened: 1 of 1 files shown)


0xTim:
This looks like a hangover from the Swift 3 days. If you're converting the `Package.swift` to Swift 4 (which is something that should definitely be done!), you'll need to add the `// swift-tools-version:4.0` at the top

2.0/docs/auth/package.md
```diff
@@ -15,7 +17,11 @@ let package = Package(
         .Package(url: "https://github.com/vapor/vapor.git", majorVersion: 2),
         .Package(url: "https://github.com/vapor/auth-provider.git", majorVersion: 1)
```
0xTim: These lines will need to change if we are moving to Swift 4 - @tanner0101 should we provide Swift 4 only or both Swift 3 and Swift 4?

vzsg: Why not keep both syntaxes?

PR #266, opened 2017-12-22
Fix incorrect auth documentation

revolter:


2.0/docs/auth/provider.md
```diff
@@ -4,10 +4,10 @@ After you've [added the Auth Provider package](package.md) to your project, sett
 
 ## Add to Droplet
 
-Register the `AuthProvider.Provider` with your Droplet.
+Register the `AuthProvider.Provider` with your Droplet in your `main.swift` file:
```
0xTim: Generally we don't want people to touch their `main.swift`, it should all be done in `Config+Setup.swift` for Vapor 2

revolter: Yeah, I figured that out after creating this.

PR #286, opened 2018-02-21
LeafRenderer has no member make()

jonny7:
Looking through the class, possibly make() has been replaced by render()


0xTim:
@jonny7 And so it begins! Thanks for these - the docs are majorly out of date due to the betas evolving too quickly to make it worth while keeping up. There will be a big documentation write up when the RC hits and the major bugs are fixed hopefully starting next week I should guess. But you are right, it is `render` now!

jonny7:
@0xTim No problem, nice to help out  :-)

PR #304, opened 2018-04-15
Specify rc.2 for FluentPostgreSQL package

adtrevor:
The first rc was not compatible with vapor-rc.2


adtrevor:
That's right, actually the issue I had din't come from that. Thank you !

PR #323, opened 2018-05-12
Structs with Futures are only Encodable

tiferrei:

PR #324, opened 2018-05-12
Futures can't be Decodable!

0xTim:
As it says on the tin, fix the docs as it has caused confusion!


0xTim:
Closing in favour of #323

PR #368, opened 2018-10-19
Add bool type support documentation #27

vi4m:


grundoon:
For completeness, this is missing a modified example in the `Form` section.

vi4m:
I did it on purpose. Html widget for real-live boolean type would require additional JS code. 

Using just: 
```
     <input type="text" name="name">
     <input type="text" name="age">
     <input type="file" name="image">
+    <input type="checkbox" name="isAdmin" value="true">
```
wouldn't work, because  `isAdmin` field will be sent only when checked("true").  When not checked, value will not be sent at all, and field wouldn't be present for multipart decoding.

PR #370, opened 2018-10-24
Update style-guide.md

### LINKED ISSUE #369: [Styleguide] Swift 4.2 fails with API Controllers folder
Swift 4.2 fails to compile in Linux when you have a folder `API Controllers`

gperdomor:
Fix #369


grundoon:
Might be a good idea to also document *why* (for the time being, anyway) this is a needed best practice.

mcdappdev:
Yeah, I think that'd be a good idea.

mcdappdev:
Fixed by #372

PR #371, opened 2018-10-30
Update models.md to add `Timestampable` section

axtonpitt:
It was difficult to find documentation on how to use `Timestampable`, as it seems similar to Vapor 2 maybe that guide can be used.

(Was not sure which base branch to target as well.)

### DIFF
--- a/3.0/docs/fluent/models.md
+++ b/3.0/docs/fluent/models.md
@@ -177,6 +177,39 @@ final class User: Model {
 }
 ```
 
+## Timestampable
+
+To add timestamps to your model, simply conform it to `Timestampable`.
+
+```swift
+extension User: Timestampable { }
+```
+
+You can access the updated at and created at times on any model instance.
+
+```swift
+user.updatedAt // Date?
+user.createdAt // Date?
+```
+
+When filtering or sorting on the timestamp data, you can use the timestamp keys from the class.
+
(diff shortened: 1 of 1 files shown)


grundoon:
Hi @axtonpitt – the `Timestampable` protocol was essentially merged into `Model` for Vapor 3. It no longer exists as a standalone which requires explicit conformance adoption. Rather, it is merely the (optional) presence of the required keys in a model that triggers the functionality, as documented in https://api.vapor.codes/fluent/latest/Fluent/Protocols/Model.html

calebkleveter:
That's true, but I think documenting how to integrate timestamping is a good idea. I think we just need to remove the part about the `Timestampable` protocol.

mcdappdev:
Hi @axtonpitt - let me know if you want to move forward with this. If so, we'll want to address the comments left on this thread, specifically the removal of the `Timestampable` mention. Thanks!

axtonpitt:
Hey @mcdappdev, yeah keen to move forward. Should I keep this PR just to the timestamping properties?

mcdappdev:
Yeah I think that’d be perfect! Thanks again :)

axtonpitt:
Sorry have neglected this PR 😅, #403 looked great and addresses this issue. Closing.

PR #372, opened 2018-11-08
Update style-guide.md

steffendsommer:
I'm not sure about the underlying issue, but we've had issues with spaces in filepaths in Swift 4.2 on Linux. In general, I prefer to avoid spaces as it tends to lead to trouble. Also removing the spaces increases the consistency in the styleguide with regards to names in files and folders (and makes it easier to understand if or when to use spaces).

What do you think?


penny-coin:
Hey @steffendsommer, you just merged a pull request, have a coin! 

You now have 9 coins.

PR #403, opened 2019-05-04
Add documentation for createdAt & updatedAt timestamps and for soft delete

Jerry-Carter:
Added descriptions of the creation and update timestamps and of soft delete to the models section of Fluent.

### DIFF
--- a/3.0/docs/fluent/models.md
+++ b/3.0/docs/fluent/models.md
@@ -232,6 +232,57 @@ let didDelete = user.delete(on: req)
 print(didDelete) /// Future<Void>
 ```
 
+## Creation and Last Update Timestamps
+
+Often an application requires knowing when a row / item is created or gets updated.  If the model specifies fields for storing creation and/or update timestamps, Fluent will automatically set these values.  The model simply specifies the name of the field and a matching definition.
+
+```swift
+static let createdAtKey: TimestampKey? = \.createdAt
+static let updatedAtKey: Timestampkey? = \.updatedAt
+
+var createdAt: Date?
+var updatedAt: Date?
+```
+
+The presence of `createdAtKey` causes the associated field to be set when the object is created.  This field will not be subsequently changed.  The presence of `updatedAtKey` causes the associated field to be set when the object is created and whenever any update operation is performed.  A model may define one, both, or neither.
+
+
+## Soft Delete
+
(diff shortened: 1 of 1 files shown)


Jerry-Carter:
This might also address <https://github.com/vapor/documentation/pull/371>

3.0/docs/fluent/models.md
```diff
@@ -232,6 +232,57 @@ let didDelete = user.delete(on: req)
...
+
+The restore will clear the deletion timestamp and will also set the update timestamp.  The restored rows will now appear in queries as normal.
+
+Rows, both normal and soft deleted, may be permanently deleted by indicating `force: true` on the delete.
+
+```swift
+User.query(on: conn, withSoftDeleted:true).filter(…).delete(force: true)
+```
+
+!!! tip
```
mcdappdev: I’m not sure I understand the purpose or intention behind this tip. Maybe I just don't get the usecase.

Jerry-Carter: This is the challenge of covering material in a brief manner.  All the other sections in the document are short so I tried to do the same.  I could add a longer discussion of the common use cases for soft delete and some of corresponding issues.

The two most common use cases for soft delete, from my experience, are (1) archiving of deleted rows and (2) undo functionality.

In the second case, maybe there is a table of messages which the user can review and delete, but there is a desire to allow the user to change their mind for a period of time.  Because this table has no dependencies, the restore operation is safe.  There is typically a batch process that grabs all the soft deleted rows after maybe two weeks and removes them from the database.

In the first case, which the tip covers, the business rules or legal constrains may require that no rows ever be removed.  Here soft delete is often used for expedience, but often results in a large table with few active rows.  There are two preferred solutions.  Either use a trigger or similar logic (e.g. willDelete) to transfer these to a separate table containing only the deleted rows.  This reduces the size of the active table and ensures that the database level constraints such as uniqueness and foreign keys apply only to the currently active rows.  The second case is to use a batch process, perhaps nightly, to grab the soft deleted rows and transfer them to the table of deleted data.  The tip is intended to highlight this case and remind the reader of the lifecycle event.  They may chose to use a batch process or database trigger instead.

Jerry-Carter: For completeness, the use case in my current application is different.  Deletes are rare and significant when they do occur.  Here soft delete is used to deprecate a row.  The 'willDelete' function is used to delete most of the rows with a foreign key relation.  Other tables require manual row-by-row decisions for migration.  Soft delete here prevents new uses of the deprecated row while also allowing time for propagating the impact through the rest of the graph.

mcdappdev: Ah ok, thanks for clarifying! That makes more sense.

penny-coin:
Hey @Jerry-Carter, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #407, opened 2019-06-12
Cannot use `id` in Swift

Yasumoto:
Prevent this when doing a direct copy-paste:

`error: 'id' is unavailable in Swift: 'id' is not available in Swift; use 'Any'`

Found some details on the [Swift blog](https://developer.apple.com/swift/blog/?id=39)

### DIFF
--- a/3.0/docs/getting-started/routing.md
+++ b/3.0/docs/getting-started/routing.md
@@ -40,7 +40,7 @@ public func routes(_ router: Router) throws {
     router.get("hello") { req in
         return "Hello, world!"
     }
-  
+
     /// ...
 }
 ```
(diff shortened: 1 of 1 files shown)


Yasumoto:
Actually I mixed up where the error message was happening since I pulled out the code using this into a separate function 😅 . That said, I'd suggest we move away from it in general so we get better error messages, but I defer to y'all!

0xTim:
I think it's probably ok. If we want to change it, I prefer `userID` over `uid`

Yasumoto:
Friendly ping!

Yasumoto:
😂  Thanks, @0xTim ! 🎉

PR #425, opened 2019-12-20
Add unauthenticated user redirection example

Kireyin:


3.0/docs/auth/web.md
```diff
@@ -99,4 +99,19 @@ Upon visiting `/hello`, you should recieve an error message stating that you are
...
 If you open the inspector, you should notice a new cookie named `"vapor-session"` has been added to your browser.
 
+### Redirecting unauthenticated users to the Login page
 
+To redirect unauthenticated users from protected routes to the login page, add the `RedirectMiddleware` to your protected routes.
+
+```swift
+func boot(router Router) throws {
+    let auth = User.authSessionsMiddleware() // The Authentication Middleware
+    let redirect = RedirectMiddleware<User>(path: "login") // The redirection Middleware
```
0xTim: Could you change the comment here to say what `login` is - something like "Create the redirect middleware providing the path to redirect to if the user is not logged in"?

Kireyin: Yes 👍🏻

0xTim:
Thanks!

PR #436, opened 2020-01-27
Timestamp and ModelMiddleware docs

brycesteve:
Timestamp/Soft Delete docs added and info on using ModelMiddleware

Also added v4 to web.Dockerfile for local testing.

### DIFF
--- a/4.0/docs/fluent/overview.md
+++ b/4.0/docs/fluent/overview.md
@@ -88,7 +88,7 @@ Using convenience inits is especially helpful if you add new properties to your 
 
 ### Timestamps
 
-If your model needs to track when it is created or gets updated. You can specify fields for storing creation and/or update timestamps, Fluent will automatically set these values. 
+Fluent provides the ability to track creation and update times on models by specifying `Timestamp` fields in your model. Fluent automatically sets the fields when necessary. You can add these like so:
 
 ```swift
 @Timestamp(key: "created_at", on: .create)
@@ -99,10 +99,7 @@ var updatedAt: Date?
 ```
 
 !!! Info
-    You can use any name/key for these fields. created_at/updated_at, are only for illustration purposes
-
-If you use timestamps, Fluent will take care of setting them for you when your model is updated of deleted.
-
+    You can use any name/key for these fields. `created_at` / `updated_at`, are only for illustration purposes
 
 
 ## Migrations
@@ -116,6 +113,8 @@ struct CreateGalaxy: Migration {
         database.schema("galaxies")
             .field("id", .int, .identifier(auto: true))
             .field("name", .string)
+            // These are necessary if you have added timestamps
+            //in your models
             .field("created_at", .datetime)
             .field("updated_at", .datetime)
             .create()
@@ -407,9 +406,9 @@ A key-path to the `@Children` relation is passed to `with` to tell Fluent to aut
 
 ## Lifecycle
 
-To create hooks that respond to Events on your Model, you can create Middlewares for your model. Middleware must conform to ModelMiddleware.
+To create hooks that respond to events on your Model, you can create middlewares for your model. Your middleware must conform to `ModelMiddleware`.
 
-Here is an example of a simple Middleware:
+Here is an example of a simple middleware:
 
 ```swift
 struct GalaxyMiddleware: ModelMiddleware {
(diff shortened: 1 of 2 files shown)


4.0/docs/fluent/overview.md
```diff
@@ -384,6 +405,107 @@ A key-path to the `@Children` relation is passed to `with` to tell Fluent to aut
...
+
+```swift
+app.databases.middleware.use(GalaxyMiddleware(), on: .psql)
+```
+
+## Soft Delete
+
+Soft deleting an item sets a timestamp on it to the time when it was soft deleted, but doesn't actually remove it from the database. Soft deleted items won't appear in queries by default, and can be restored at any time.
+
+In the same way as created/deleted timestamps, to enable soft deleting items, all that is required is to add a timestamp property, that responds to the .delete trigger.
```
0xTim: Nit, clean up:
> Similar to created and deleted timestamps, to enable soft deletion in a model just set a deletion timestamp for `.delete`:

4.0/docs/fluent/overview.md
```diff
@@ -384,6 +405,107 @@ A key-path to the `@Children` relation is passed to `with` to tell Fluent to aut
...
+
+```swift
+@Timestamp(key: "deleted_at", on: .delete)
+    var deletedAt: Date?
+```
+
+Calling `Model.delete(on: Database)` on a model that has a delete timestamp property will automatically soft delete. Without this property, the item will be permanently deleted.
```
0xTim: Nit - change the function signature to `Model.delete(on:)`

Nit - change the end of the first sentence to `will automatically soft delete it`. Remove the second sentence.

4.0/docs/fluent/overview.md
```diff
@@ -384,6 +405,107 @@ A key-path to the `@Children` relation is passed to `with` to tell Fluent to aut
...
+
+If you need to perform a query that includes the soft deleted items, you can use `withDeleted()`. Soft deleted items can easily then be restored with `restore(on: database)`:
```
0xTim: Change to
> you can use `withDeleted()` in your query.

Remove the passive tense in the second sentence:
> You can easily restore soft deleted items with `restore(on:)`:

4.0/docs/fluent/overview.md
```diff
@@ -86,6 +86,25 @@ init(id: Int? = nil, name: String) {
...
+```swift
+@Timestamp(key: "created_at", on: .create)
+var createdAt: Date?
+    
+@Timestamp(key: "updated_at", on: .update)
+var updatedAt: Date?
+```
+
+!!! Info
+    You can use any name/key for these fields. created_at/updated_at, are only for illustration purposes
```
0xTim: Change it so the keys are monospace: "`created_at`/`updated_at`"

4.0/docs/fluent/overview.md
```diff
@@ -384,6 +405,107 @@ A key-path to the `@Children` relation is passed to `with` to tell Fluent to aut
...
+```swift
+// Find the first soft deleted Galaxy
+Galaxy.query(on: db).withDeleted().first().flatMap{ g -> EventLoopFuture<Void> in
+    guard g != nil else { return db.eventLoop.future() }
+    // Restore galaxy
+    return g.restore(on: db)
+}
+```
+
+It is also equally easy to permanently delete items:
```
0xTim: Expand by saying "delete items with the `force` parameter:"

4.0/docs/fluent/overview.md
```diff
@@ -384,6 +405,107 @@ A key-path to the `@Children` relation is passed to `with` to tell Fluent to aut
...
+        return next.softDelete(model, on: db)
+    }
+    
+    // Runs when a soft deleted model is restored
+    func restore(model: Galaxy, on db: Database, next: AnyModelResponder) -> EventLoopFuture<Void> {
+        return next.restore(model , on: db)
+    }
+    
+    // Runs when a model is permanently deleted
+    // The "force" parameter indicates that the model will
```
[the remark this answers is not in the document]
brycesteve: @0xTim You might be correct about this.  I wasn't completely sure about the order that things were called in this code:

```swift
public func delete(force: Bool = false, on database: Database) -> EventLoopFuture<Void> {
        if !force, let timestamp = self.timestamps.filter({ $0.1.trigger == .delete }).first {
            timestamp.1.touch()
            return database.configuration.middleware.chainingTo(Self.self) { event, model, db in
                model.handle(event, on: db)
            }.handle(.softDelete, self, on: database)
        } else {
            return database.configuration.middleware.chainingTo(Self.self) { event, model, db in
                model.handle(event, on: db)
            }.handle(.delete(force), self, on: database)
        }
    }
```
from Fluent Model+CRUD.swift.  Wll wait to see if @tanner0101 can shed some light on it.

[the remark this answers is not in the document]
0xTim: Yeah looking at that code, if force is set and there's a timestamp field for `.delete` it call down to `.softDelete`, otherwise it just calls `.delete`. So looks like the comment need to be updated

4.0/docs/fluent/overview.md
```diff
@@ -384,6 +405,107 @@ A key-path to the `@Children` relation is passed to `with` to tell Fluent to aut
...
+    func delete(model: Galaxy, force: Bool, on db: Database, next: AnyModelResponder) -> EventLoopFuture<Void> {
+        return next.delete(model, force: force, on: db)
+    }
+}
+```
+
+Each of these methods has a default implementation, so you only need to include the methods for hooks that you need. You should normally return the appropriate method on the 'next' responder, so Fluent knows to continue processing the event. 
+
+!!! Important
+    The 'model' parameter in your Middleware methods should be the Type of the Model the Middleware should respond to. In the above example GalaxyMiddleware will respond to events on the Galaxy Model.
```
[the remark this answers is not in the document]
brycesteve: The ModelMiddleware protocol uses an associatedtype for Model.  I haven't tried it, but I would imagine that using different model types in the same protocol would break the associated type.

brycesteve:
Readability changes added. 

This is my first attempt at writing docs, so feel free to suggest any further improvements.

brycesteve:
Tweaks made

PR #440, opened 2020-02-08
run supervisor as vapor

EricWVGG:
When following the instructions at /docs/4.0/docs/deploy/supervisor.md, the user will encounter the following error when running `supervisorctl reread`:
```
error: <class 'socket.error'>, [Errno 13] Permission denied: file: /usr/lib/python2.7/socket.py line: 228
```
This is due to the "vapor" user not having the necessary permissions to run supervisorctl. Directions for enabling these permissions are outlined here: `Supervisor/supervisor#173`

Docs updated to guide user through giving permissions to `vapor` user.

### DIFF
--- a/4.0/docs/deploy/supervisor.md
+++ b/4.0/docs/deploy/supervisor.md
@@ -9,6 +9,28 @@ sudo apt-get update
 sudo apt-get install supervisor
 ```
 
+Supervisor can currently only run as the `root` user. You can exit and login as root, or create a new group and add `vapor` to it.
+
+```sh
+groupadd supervisor
+usermod -a vapor -G supervisor
+```
+
+Log out, log back in, and edit the the supervisord configuration file (/etc/supervisor/supervisor.conf) unix_http_server section:
+
+```sh
+[unix_http_server]
+file=/var/run/supervisor.sock ; (the path to the socket file)
+chmod=0770 ; socket file mode (default 0700)
+chown=root:supervisor
+```
+
+Finally, restart Supervisor:
+
+```sh
+supervisorctl reload
+```
+
 ## Configure
 
 Each Vapor app on your server should have its own configuration file. For an example `Hello` project, the configuration file would be located at `/etc/supervisor/conf.d/hello.conf`
--- a/4.0/docs/deploy/supervisor.md
+++ b/4.0/docs/deploy/supervisor.md
@@ -9,14 +9,14 @@ sudo apt-get update
 sudo apt-get install supervisor
 ```
 
-Supervisor can currently only run as the `root` user. You can exit and login as root, or create a new group and add `vapor` to it.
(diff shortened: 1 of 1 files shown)


4.0/docs/deploy/supervisor.md
```diff
@@ -9,6 +9,28 @@ sudo apt-get update
 sudo apt-get install supervisor
 ```
 
+Supervisor can currently only run as the `root` user. You can exit and login as root, or create a new group and add `vapor` to it.
```
0xTim: I think by default we shouldn't even be suggesting running anything as root. So I'd change this line to:

Once installed, Supervisor is only configured to run as the `root` user. The best way to improve this is to create a new group that can supervisor can use and add the `vapor` user to it:

EricWVGG:
hm, but the Digital Ocean deployment steps are pretty explicit about not running as root: https://github.com/vapor/docs/blob/master/4.0/docs/deploy/digital-ocean.md#add-user

0xTim:
Not sure I understand what you mean? Aren't we both saying we should do everything we can to discourage users from running as root?

EricWVGG:
i misread your response. i will amend the text in a bit, thanks. 

/thumbs

4.0/docs/deploy/supervisor.md
```diff
@@ -9,6 +9,28 @@ sudo apt-get update
...
+usermod -a vapor -G supervisor
+```
+
+Log out, log back in, and edit the the supervisord configuration file (/etc/supervisor/supervisor.conf) unix_http_server section:
+
+```sh
+[unix_http_server]
+file=/var/run/supervisor.sock ; (the path to the socket file)
+chmod=0770 ; socket file mode (default 0700)
+chown=root:supervisor
```
tanner0101: What does this change to supervisor's configuration do? I don't see any mention of the newly created `vapor` user here.

EricWVGG: The `vapor` user was added to a newly created `supervisor` group earlier.
https://github.com/vapor/docs/pull/440/files#diff-f5d69c42e86da802420707018a08b9d4R16

This bit alters Supervisor to run off a group `supervisor`. explanation here:
https://github.com/Supervisor/supervisor/issues/173#issuecomment-186128727

0xTim:
LGTM - will wait for Tanner to OK his comments

4.0/docs/deploy/supervisor.md
tanner0101: Ok, that makes more sense. Although I'm still confused how exactly you tell supervisor to run as the `vapor` user or under the `supervisor` group. 

Is that working because you're calling `supervisorctl reload` without `sudo` later?

0xTim:
@EricWVGG are you still wanting to merge this?

0xTim:
I'm going to close this as it's out of date and been a while. Feel free to reopen if you want to add it!

PR #478, opened 2020-04-30
Fix async example to use runIfActive instead of submit

finestructure:
As discussed here: https://discordapp.com/channels/431917998102675485/444249946808647699/705452538438221845

`@jw` advises *not* to use `submit`:

```
don't use submit really. It calls you back with .cancelled if the thread pool is already torn down
```

https://discordapp.com/channels/431917998102675485/444249946808647699/705420024592072765

### DIFF
--- a/4.0/docs/async.md
+++ b/4.0/docs/async.md
@@ -287,23 +287,16 @@ Make sure to run any blocking work in the background. Use promises to notify the
 
 ```swift
 router.get("hello") { req -> EventLoopFuture<String> in
-    /// Create a new void promise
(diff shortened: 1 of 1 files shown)


0xTim:
We'll need to update the async password hasher to use this new method as well, I'll get around to that tomorrow and then merge this PR

mcdappdev:
@0xTim I can knock this out today, if you'd like.

mcdappdev:
@0xTim PR here: https://github.com/vapor/vapor/compare/async-password?expand=1

PR #483, opened 2020-05-08
Update queues.md

Andrewangeta:
Adds the queues mongo driver to the community drivers.


Andrewangeta:
Sure. Makes sense to me.

PR #484, opened 2020-05-08
Docs 4.0: Corrects ngnix example config to use 'location /' instead of 'location @proxy'

vpsteinmann:
The `@proxy` location was not defined, so the the server returned an `403 Forbidden` error. 
Use `location /` instead.

### DIFF
--- a/4.0/docs/deploy/nginx.md
+++ b/4.0/docs/deploy/nginx.md
@@ -82,7 +82,7 @@ server {
 
     root /home/vapor/Hello/Public/;
 
-    location @proxy {
+    location / {
         proxy_pass http://127.0.0.1:8080;
         proxy_pass_header Server;
         proxy_set_header Host $host;


0xTim:
I'm not sure what is correct, I'm guessing it should be `/` or `@proxy` with the `try_files` directive set. That's what it seems to be missing compared to the RW book that uses `@proxy` and I'm not sure what's better?

vpsteinmann:
That's right. Either just `/` or `@proxy` with the `try_files` directive. But since the `try_files` directive is explained in the next paragraph, I thought just `/` would be simpler. 

But we could also use this snippet that incorporates both:

```
server {
    server_name hello.com;
    listen 80;

    root /home/vapor/Hello/Public/;

    location / {
        try_files $uri @proxy;
    }

    location @proxy {
        proxy_pass http://localhost:8080;
        proxy_pass_header Server;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_pass_header Server;
        proxy_connect_timeout 3s;
        proxy_read_timeout 10s;
    }
}
```

0xTim:
Ah ok, so in that case I suggest we switch the first example to `/` as per your PR and just clarify the `try_files` bit (it can be a bit confusing about what to replace) to make it clear

PR #490, opened 2020-05-20
Remove reference to userInfo

siemensikkema:


4.0/docs/upgrading.md
```diff
@@ -314,8 +314,6 @@ app.myNumber = 42
 print(app.myNumber) // 42
 ```
 
-Both Application and Request also have `userInfo` dictionaries for storing any data you may need.
```
gwynne, line 317: I think this should be updated to refer to the `storage` mechanism that replaced `userInfo` rather than completely removed, assuming `Storage` isn't already mentioned elsewhere on the page. (Also, `Application` and `Request` should have backticks to give them `&lt;code&gt;` style.

siemensikkema: @gwynne I opted to simply remove it since the storage mechanism is referred to in https://docs.vapor.codes/4.0/upgrading/#custom-providers.
Since I only removed a sentence, I don't think adding backticks is relevant? 🤔

gwynne: Ah, okay, if it's already being referred to elsewhere then that's fine 🙂 And yeah, the backticks would only matter if you needed to add a comment about Storage, but since you don't, it's all good as is 🙂

PR #492, opened 2020-05-20
Improve wording in throwing map / flatMap section

siemensikkema:


4.0/docs/upgrading.md
```diff
@@ -392,7 +392,7 @@ futureA.flatMap { a in
 }
 ```
 
-Flat-maps that _do_ throw must return a future error.
+Flat-maps that _do_ throw must now instead return a future error.
```
0xTim, line 395: We should probably make it explicit here that the error must be handled and caught inside the `flatMap`

siemensikkema: @0xTim That's what the next example is about. This is is about explicitly throwing an error in the closure. I also think it is confusing but I wasn't sure how to resolve it. Maybe we can merge them like this:
> Instead of throwing an error inside a flat-map return a future error. If the error originates from another throwing method, the error can be caught in a do / catch and returned as a future.
followed by the example from line 410. I think that should give the reader enough to understand it, no?

0xTim: Yeah I think that's a good compromise

PR #495, opened 2020-05-21
Add controllers section.

ileitch:


4.0/docs/routing.md
```diff
@@ -296,6 +296,31 @@ app.get("hello", ":name") { req in
 }.description("says hello")
 ```
 
+### Controllers
+
+The controller pattern is a great way to keep your routes file small by separating route handlers out into distinct structs.
+
+```swift
+struct TodosController {
```
0xTim, line 304: I think any controller should default to `RouteCollection`s and all docs should show that as the preferred way.

0xTim:
To add to the above comment, I think controller should probably go in their own file a la https://docs.vapor.codes/3.0/getting-started/controllers/

There's a lot to talk about with Controllers so it's better to separate that out that polluting the routes docs

4.0/docs/controllers.md
```diff
@@ -0,0 +1,60 @@
...
+	func index(req: Request) throws -> String {
+		...
+	}
+
+    func create(req: Request) throws -> String {
+        ...
+    }
+
+    func show(req: Request) throws -> String {
+        let id = req.parameters.get("id")!
```
0xTim, line 24: I don't think anything in the docs should be force unwrapping IDs, people _will_ just copy that blindly

```suggestion
        guard let id = req.parameters.get("id") else {
            throw Abort(.internalServerError
        }
        // ...
```

4.0/docs/controllers.md
```diff
@@ -0,0 +1,60 @@
...
+        ...
+    }
+
+    func update(req: Request) throws -> String {
+        let id = req.parameters.get("id")!
+        ...
```
0xTim, line 30: As above

```suggestion
        guard let id = req.parameters.get("id") else {
            throw Abort(.internalServerError)
        }
        // ...
```

4.0/docs/controllers.md
```diff
@@ -0,0 +1,60 @@
...
+		...
+	}
+
+    func create(req: Request) throws -> String {
+        ...
+    }
+
+    func show(req: Request) throws -> String {
+        let id = req.parameters.get("id")!
+        ...
```
0xTim, line 25: I don't think we should be force unwrapping anything in the docs. People _will_ just copy that blindly unfortunately.

```suggestion
        guard let id = req.parameters.get("id") else {
            throw Abort(.internalServerError)
        }
        // ...
```

4.0/docs/controllers.md
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    func update(req: Request) throws -> String {
+        let id = req.parameters.get("id")!
+        ...
+    }
+
+    func delete(req: Request) throws -> String {
+        let id = req.parameters.get("id")!
+        ...
```
0xTim, line 35: As above

```suggestion
        guard let id = req.parameters.get("id") else {
            throw Abort(.internalServerError)
        }
        // ...
        ...
```

4.0/docs/controllers.md
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+```
+
+Controller methods should always accept a `Request` and return something `ResponseEncodable`.
+
+!!! note
+	[EventLoopFuture](async.md) whose expectation is `ResponseEncodable` (i.e, `EventLoopFuture<String>`) is also `ResponseEncodable`.
+
+To use this controller, we can simply initialize it, then pass the methods to the router.
```
0xTim, line 45: I'd still prefer the default way to be a `Controller` that conforms to `RouteCollection` and we register that on the routes rather than individual methods

ileitch: Can you give an example? I'm not sure what you mean.

0xTim: [Here's a controller](https://github.com/0xTim/VaporAuthExample/blob/master/Sources/App/Controllers/UserController.swift#L6) that conforms to `RouteCollection`. You basically register all the routes on `boot(routes:)`. This keeps the knowledge of the routes (like what parameters it needs in that file) and helps split things up so you don't end up with a massive routes.swift. You then register the controller [like so](https://github.com/0xTim/VaporAuthExample/blob/master/Sources/App/routes.swift#L13), instead of having to register each route individually in routes.swift.

0xTim: The vapor/template also needs updating, I haven't gotten around to that (it should have never not been a `RouteCollection` in the first place

PR #505, opened 2020-06-07
Docs 4.0: Create heroku.md file

ggrana:
This is basically the same file from the 3.0 docs, which worked well to me to start the heroku app. Just did the needed changes to work with 4.0.

### DIFF
--- a/4.0/docs/deploy/heroku.md
+++ b/4.0/docs/deploy/heroku.md
@@ -1,0 +1,228 @@
+# What is Heroku
+
+Heroku is a popular all in one hosting solution, you can find more at heroku.com
+
+## Signing Up
+
+You'll need a heroku account, if you don't have one, please sign up here: [https://signup.heroku.com/](https://signup.heroku.com/)
+
+## Installing CLI
+
+Make sure that you've installed the heroku cli tool.
+
+### HomeBrew
+
+```bash
+brew install heroku/brew/heroku
+```
+
+### Other Install Options
+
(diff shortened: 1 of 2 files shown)


4.0/docs/deploy/heroku.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+**DATABASE_URL** here will represent out postgres database. **NEVER** hard code the static url from this, heroku will rotate it and it will break your application. It is also bad practice.
+
+Here is an example databsae configuration
+
+```swift
+    app.databases.use(try .postgres(
+        url: Environment.get("DATABASE_URL") ?? "some_fallback_url_if_needed"
+    ), as: .psql)
```
0xTim, line 202: We should probably avoid pointing people to a fallback URL and change it around to match the 3.0 guide:

```suggestion
if let databaseURL = Environment.get("DATABASE_URL") {
    app.databases.use(try .postgres(
        url: databaseURL
    ), as: .psql)
} else {
    // ...
}
```

ggrana: I am following the format that Vapor creates the project, which looks like this:

```
    app.databases.use(.postgres(
        hostname: Environment.get("DATABASE_HOST") ?? "localhost",
        username: Environment.get("DATABASE_USERNAME") ?? "username",
        password: Environment.get("DATABASE_PASSWORD") ?? "password",
        database: Environment.get("DATABASE_NAME") ?? "database"
    ), as: .psql)
```

ggrana: But your option looks good to me!

PR #513, opened 2020-06-18
Upgrade to 4: Example of Fluent save DOES not return

hwde:


0xTim:
The docs are correct - if your route returns the model after a save in Vapor 3 you would do
```swift
return model.save(on: req)
```

That would return `EventLoopFuture<Void>` now, so you need to add the map to get it to return the model that's been saved (using reference semantics)

PR #516, opened 2020-06-29
Added mutating keyword to afterDecode and beforeEncode hooks

iComputerfreak:
The mutating keyword is required, as both functions assign a new value to `self.name`.
The current code example results in the following error: **Cannot assign to property: 'self' is immutable**

### DIFF
--- a/4.0/docs/content.md
+++ b/4.0/docs/content.md
@@ -126,7 +126,7 @@ Vapor will automatically call `beforeDecode` and `afterDecode` on a `Content` ty
 
 ```swift
 // Runs after this Content is decoded.
-func afterDecode() throws {
+mutating func afterDecode() throws {
     // Name may not be passed in, but if it is, then it can't be an empty string.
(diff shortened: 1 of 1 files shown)


4.0/docs/content.md
```diff
@@ -126,7 +126,7 @@ Vapor will automatically call `beforeDecode` and `afterDecode` on a `Content` ty
 
 ```swift
 // Runs after this Content is decoded.
-func afterDecode() throws {
+mutating func afterDecode() throws {
```
0xTim, line 129: You're correct, this is required because the docs use a `struct` - could you also add to the comment above (and in the one for `beforeEncode()` to say `mutating` is required for `strut`s and not `class`

iComputerfreak: Not sure, if that is necessary. If you use the `mutating` keyword inside a class, Xcode will tell you to remove it.

0xTim: That is true, yes, but we don't want people to be copying code across and immediately getting a warning - the docs should help them as much as they can

PR #529, opened 2020-07-26
Update references to `app.swift`

tylermilner:
When performing `vapor new hello -n` command as part of [Getting Started](https://docs.vapor.codes/4.0/hello-world/), an `app.swift` file is not generated. The code that creates and runs the `app` lives in `main.swift`. This updates the documentation to reflect this.

Reference file structure for `vapor new hello -n` command:
image

### DIFF
--- a/4.0/docs/folder-structure.md
+++ b/4.0/docs/folder-structure.md
@@ -10,7 +10,6 @@ Now that you've created, built, and run your first Vapor app, let's take a momen
 │   │   ├── Controllers
 │   │   ├── Migrations
 │   │   ├── Models
-│   │   ├── app.swift
 │   │   ├── configure.swift
 │   │   └── routes.swift
 │   └── Run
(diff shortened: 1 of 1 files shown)


4.0/docs/folder-structure.md
```diff
@@ -58,13 +57,13 @@ The migrations folder is where your database migrations go if you are using Flue
 
 The models folder is a great place to store your `Content` structs or Fluent `Model`s.
 
-#### app.swift
+#### main.swift
```
0xTim, line 60: I think the main.swift file should live a few levels up, otherwise this could make it appear that `main.swift` live in the `App` directory in `Sources

tylermilner: I created a "Run" folder and moved the reference to `main.swift` accordingly so that it better matches the structure on disk. Let me know if you want any wording tweaks.

PR #543, opened 2020-08-26
Bumped fluent-sqlite-driver version to 4.0.0 in Fluent Overview

bdrelling:
I was going through this and copy+pasted, then noticed it still had the `-beta` tag. Wanted to bump this just in case anyone else makes the same mistake!

I bumped it to `4.0.0` after checking the [releases for vapor/fluent-sqlite-driver](https://github.com/vapor/fluent-sqlite-driver/releases). I see there's also a `4.0.1`, but I feel like `4.0.0` was more appropriate.


0xTim (approved):
Thanks!

PR #547, opened 2020-09-24
Added example of how to configure Postgres database with unverified TLS for Heroku deployment

tylermilner:
I noticed the [3.0 docs](https://docs.vapor.codes/3.0/deploy/heroku/#configure-the-database) showed how to configure unverified TLS for your Postgres database when deploying to Heroku. It took me a little while to figure out the right way to do this in Vapor 4 so I figured it might be worthwhile to include it in the 4.0 docs.


0xTim (approved):
Thanks!

PR #550, opened 2020-09-29
SQLite foreign key violation tip

Craz1k0ek:
I ran into an issue where `PRAGMA foreign_keys = ON` was required on the database itself. I was unaware of this, as it worked for Vapor. I added a tip containing problems that can occur when querying the database directly.


0xTim (approved):
Thanks!

PR #552, opened 2020-10-01
Add example to use MySQL in local without SSL

Lloople:
Faced this issue when starting a new project, found the answer [here](https://forums.swift.org/t/fluent-connect-mysql-8-failure-how-to-solve/35939/14) and applied it [here](https://github.com/Lloople/where-is-my-stuff/blob/master/Sources/App/configure.swift#L21). 

It's working as expected so I think for new users would be useful to have this information directly in vapor docs 👍

### DIFF
--- a/4.0/docs/fluent/overview.md
+++ b/4.0/docs/fluent/overview.md
@@ -150,6 +150,9 @@ app.databases.use(.mysql(
 ), as: .mysql)
 ```
 
+!!! warning
+    Do not disable certificate verification in production. You should provide a certificate to the `TLSConfiguration` to verify against. 
+
 #### MongoDB
 
 MongoDB is a popular schemaless NoSQL database designed for programmers. The driver supports all cloud hosting providers and self-hosted installations from version 3.4 and up.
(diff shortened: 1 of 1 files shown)


4.0/docs/fluent/overview.md
```diff
@@ -138,6 +138,18 @@ You can also parse the credentials from a database connection string.
 try app.databases.use(.mysql(url: "<connection string>"), as: .mysql)
 ```
 
+To configure a local connection without SSL certificate involved, you should disable certificate verification.
```
0xTim, line 141: I'd add a note here about why we might want to do this and also a big banner saying don't do this in production
```suggestion
To configure a local connection without SSL certificate involved, you should disable certificate verification. You might need to do this for example if connecting to a MySQL 8 database in Docker.
```

4.0/docs/fluent/overview.md
```diff
@@ -138,6 +138,18 @@ You can also parse the credentials from a database connection string.
...
+```swift
+app.databases.use(.mysql(
+    hostname: "localhost", 
+    username: "vapor", 
+    password: "vapor", 
+    database: "vapor", 
+    tlsConfiguration: .forClient(certificateVerification: .none)
+), as: .mysql)
+```
+
```
0xTim, line 152:
```suggestion

!!! warning
    Do not disable certificate verification in production. You should provide a certificate to the `TLSConfiguration` to verify against. 

```

0xTim (changes requested):
Add a production warning otherwise @gwynne will kill me

0xTim (approved):
Thanks!

PR #571, opened 2020-11-28
Update relations document

kkiermasz:
Add missing argument in "detach" method call.

Before:
`earth.$tags.detach(inhabited)`

After:
`earth.$tags.detach(inhabited, on: database)`


0xTim (approved):
Thanks!

PR #577, opened 2020-12-11
Add documentation for Route Collections

Lloople:
I think this way it's easier to understand at a glance what's going on. 

I've also added a brief reference to routes.swift file.

### DIFF
--- a/4.0/docs/routing.md
+++ b/4.0/docs/routing.md
@@ -95,7 +95,7 @@ Now that you understand the basics, check out each section to learn more about p
 
 ## Routes
 
-A route specifies a request handler for a given HTTP method and URI path. It can also store additional metadata.
+A route specifies a request handler for a given HTTP method and URI path. It can also store additional metadata. All routes are registered in `routes.swift` file.
 
 ### Methods
 
@@ -405,8 +405,11 @@ app.group(RateLimitMiddleware(requestsPerMinute: 5)) { rateLimited in
 This is especially useful for protecting subsets of your routes with different authentication middleware. 
 
 ```swift
-app.post("login") { ... }
-let auth = app.grouped(AuthMiddleware())
-auth.get("dashboard") { ... }
-auth.get("logout") { ... }
+app.post("login") { req in ... }
+
+app.group(AuthMiddleware()) { auth in
+    auth.get("dashboard") { req in ... }
+    auth.get("logout") { req in ... }
+
+}
 ```
--- a/4.0/docs/routing.md
+++ b/4.0/docs/routing.md
@@ -95,7 +95,7 @@ Now that you understand the basics, check out each section to learn more about p
 
 ## Routes
 
-A route specifies a request handler for a given HTTP method and URI path. It can also store additional metadata. All routes are registered in `routes.swift` file.
+A route specifies a request handler for a given HTTP method and URI path. It can also store additional metadata. A good starting point for your routes is the `routes.swift` file found in the root of the project. From there you can register your routes one at a time or use `RouteCollection` types to keep them organized.
 
 ### Methods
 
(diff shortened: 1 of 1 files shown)


4.0/docs/routing.md
```diff
@@ -95,7 +95,7 @@ Now that you understand the basics, check out each section to learn more about p
 
 ## Routes
 
-A route specifies a request handler for a given HTTP method and URI path. It can also store additional metadata.
+A route specifies a request handler for a given HTTP method and URI path. It can also store additional metadata. All routes are registered in `routes.swift` file.
```
0xTim, line 98: This is misleading I'd say. It's certainly possible to add all routes in routes.swift but you don't want to register every route there (since you could end up with a file that 1000s of lines long). You can register routes with a `RouteCollection` and register the `RouteCollection` in routes.swift, but you can register a `RouteCollection` inside a `RouteCollection` etc etc. This also assumes that you're using the default template, which some people might now be. I think this should be changed to "you can register routes in routes.swift"

4.0/docs/routing.md
```diff
@@ -405,8 +405,11 @@ app.group(RateLimitMiddleware(requestsPerMinute: 5)) { rateLimited in
 This is especially useful for protecting subsets of your routes with different authentication middleware. 
 
 ```swift
-app.post("login") { ... }
-let auth = app.grouped(AuthMiddleware())
-auth.get("dashboard") { ... }
-auth.get("logout") { ... }
+app.post("login") { req in ... }
+
+app.group(AuthMiddleware()) { auth in
```
0xTim, line 410: I'm not personally a fan of the nesting (since you can end up in a pyramid of doom with futures and we want to avoid any additional nesting if we can). I think we could offer both options here.

0xTim (changes requested):

4.0/docs/routing.md
jonny7: I had the same thoughts on this as you. I do think it makes it more complex rather than less complex to a newcomer

4.0/docs/routing.md
```diff
@@ -404,9 +404,80 @@ app.group(RateLimitMiddleware(requestsPerMinute: 5)) { rateLimited in
...
+
+### Route Collections
+
+`RouteCollection` is a type offered by Vapor to help you organize your routes in different files. Using the previous example:
+
+```swift
+// routes.swift
+func routes(_ app: Application) throws {
+    try app.register(collection: GuestRouteCollection())
+
```
0xTim, line 465: Nit remove empty line

4.0/docs/routing.md
```diff
@@ -404,9 +404,80 @@ app.group(RateLimitMiddleware(requestsPerMinute: 5)) { rateLimited in
...
+func routes(_ app: Application) throws {
+    try app.register(collection: GuestRouteCollection())
+    
+    try app.register(collection: AuthRouteCollection())
+}
+
+// Routes/GuestRouteCollection.swift
+struct GuestRouteCollection: RouteCollection {
+    func boot(routes: RoutesBuilder) throws {
+        routes.get("login") { req in ... }
```
0xTim, line 472: These should probably be split out into functions instead of using the closure based approach to avoid `boot(routes:)` becoming massive itself

4.0/docs/routing.md
```diff
@@ -404,9 +404,80 @@ app.group(RateLimitMiddleware(requestsPerMinute: 5)) { rateLimited in
...
+	routes.get("register") { req in ... }
```
0xTim, line 473: Nit - line spacing

4.0/docs/routing.md
```diff
@@ -404,9 +404,80 @@ app.group(RateLimitMiddleware(requestsPerMinute: 5)) { rateLimited in
...
+    auth.get("logout") { req in ... }
+
+}
+```
+
+## Organizing Routes
+
+In larger applications, your routes file could become really long, if you want to keep it simple there's a couple approaches you can use.
+
+### Methods
```
0xTim, line 432: I don't think this should be in the docs. This doesn't really solve any issue and people can work this out for themselves if they really want. Controllers should be the recommended approach

0xTim (changes requested):
I think we should be less ambiguous with the options we provide people and give them the 'best practice' as the main option

PR #578, opened 2020-12-11
Add redirection example

### LINKED ISSUE #563: Documentation is missing an example of redirecting a request.
This is for the Vapor 4 docs 👍

Lloople:
This tries to fix #563. I put it in `routing.md` for now, but I think there should be a documentation page for `Request` and another one for `Response`

### DIFF
--- a/4.0/docs/routing.md
+++ b/4.0/docs/routing.md
@@ -413,6 +413,26 @@ auth.get("logout") { ... }
 
 ## Redirections
 
+Redirects are useful in a number of scenarios, such as forwarding old locations to new ones for SEO, redirecting an unauthenticated user to the login page or maintain backwards compatibility with the new version of your API.
+
+To redirect a request, use:
+
+```swift
+req.redirect(to: "/some/new/path")
+\```
+
+You can also specify the type of redirect, for example to redirect a page permanently (for example so that your SEO is updated correctly) use:
+
+```swift
+req.redirect(to: "/some/new/path", type: .permanent)
+\```
+
+The different `RedirectType`s are:
+
+* `.permanent` - returns a **301 Permanent** redirect
+* `.normal` - returns a **303 see other** redirect. This is the default by Vapor and tells the client to follow the redirect with a **GET** request.
+* `.temporary` - returns a **307 Temporary** redirect. This tells the client to preserve the HTTP method used in the request.
+
 Sometimes you need to redirect the user from the old route path to a new one. This is necessary when you already published the website and don't want to lose any gained SEO or maintain backwards compatibility with the new version of your API.
 
 For that purpose, you can use the redirect method found in the `Request` class.
--- a/4.0/docs/routing.md
+++ b/4.0/docs/routing.md
@@ -419,13 +419,13 @@ To redirect a request, use:
 
 ```swift
 req.redirect(to: "/some/new/path")
-\```
+```
 
-You can also specify the type of redirect, for example to redirect a page permanently (for example so that your SEO is updated correctly) use:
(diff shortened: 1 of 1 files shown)


jonny7:
This is a good addition if it's missing. I think with regards to the example and it being an SEO need, you'd probably handle that at a server level with 301 redirects imo. Perhaps for people just starting out with web work, redirecting from an admin page with a non-logged in user to a login page, would be a useful example 🙂

4.0/docs/routing.md
```diff
@@ -410,3 +410,13 @@ let auth = app.grouped(AuthMiddleware())
 auth.get("dashboard") { ... }
 auth.get("logout") { ... }
 ```
+
+## Redirections
+
+Sometimes you need to redirect the user from the old route path to a new one. This is necessary when you already published the website and don't want to lose any gained SEO or maintain backwards compatibility with the new version of your API.
```
0xTim, line 416: I agree with @jonny7 - I think we need to distinguish between the different redirects and explain what they're used for. (But thanks for adding this!) Here are some suggested tweaks, what do you think? I've also tightened up the language.

Note I've had to add a `\` to the suggested changes to get them to work correctly, that will need to be removed before merging.

```suggestion
Redirects are useful in a number of scenarios, such as forwarding old locations to new ones for SEO, redirecting an unauthenticated user to the login page or maintain backwards compatibility with the new version of your API.

To redirect a request, use:

```swift
req.redirect(to: "/some/new/path")
\```

You can also specify the type of redirect, for example to redirect a page permanently (for example so that your SEO is updated correctly) use:

```swift
req.redirect(to: "/some/new/path", type: .permanent)
\```

The different `RedirectType`s are:

* `.permanent` - returns a **301 Permanent** redirect
* `.normal` - returns a **303 see other** redirect. This is the default by Vapor and tells the client to follow the redirect with a **GET** request.
* `.temporary` - returns a **307 Temporary** redirect. This tells the client to preserve the HTTP method used in the request.

Sometimes you need to redirect the user from the old route path to a new one. This is necessary when you already published the website and don't want to lose any gained SEO or maintain backwards compatibility with the new version of your API.
```

0xTim (changes requested):
Thanks for adding this! I've added some suggestions to expand on the different types of redirects you can use

Lloople:
Nice! I've also added a link to the full list of HTTP methods.

@0xTim what do you think about creating a `request.md` file right after `Routing` and move redirect explanation (and explain other aspects of the request) into its own file?

0xTim:
@Lloople if anything is should go in a `Response` page since it's to do with a response (even though the helper is on request). But I think it makes most sense in routing since that's what it's associated with

4.0/docs/routing.md
```diff
@@ -410,3 +410,33 @@ let auth = app.grouped(AuthMiddleware())
...
+* `.normal` - returns a **303 see other** redirect. This is the default by Vapor and tells the client to follow the redirect with a **GET** request.
+* `.temporary` - returns a **307 Temporary** redirect. This tells the client to preserve the HTTP method used in the request.
+
+> To choose the proper redirection status code check out [the full list](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_redirection)
+
+Sometimes you need to redirect the user from the old route path to a new one. This is necessary when you already published the website and don't want to lose any gained SEO or maintain backwards compatibility with the new version of your API. This can be done directly in your `routes` using a closure.
+
+```swift
+app.get("old") { req in req.redirect(to: "new") }
+```
```
0xTim, line 442: This old stuff needs to be deleted then we're good to go!

0xTim (changes requested):

0xTim (approved):
Thanks!

PR #580, opened 2020-12-13
Add example for manual lazy eager loading

Lloople:
I've also moved the eager load first as it's usually more used and important than the lazy eager loading


0xTim (approved):
LGTM - thanks!

PR #582, opened 2021-01-10
Fix a typo on the Leaf overview page.

Hailong:
Fix a typo on the Leaf overview page.

### DIFF
--- a/3.0/docs/leaf/overview.md
+++ b/3.0/docs/leaf/overview.md
@@ -100,7 +100,7 @@ If you want to use another tag as part of your condition, you should omit the `#
 }
 ```
 
-Just like in Swift, you can also use `else if` statement.s
+Just like in Swift, you can also use `else if` statement:
 
 ```leaf
 #if(title == "Welcome") {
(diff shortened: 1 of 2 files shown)


3.0/docs/leaf/overview.md
```diff
@@ -100,7 +100,7 @@ If you want to use another tag as part of your condition, you should omit the `#
 }
 ```
 
-Just like in Swift, you can also use `else if` statement.s
+Just like in Swift, you can also use `else if` statement:
```
siemensikkema, line 103: Good catch! I think the proper fix would contain the plural `s` though, like so:

```suggestion
Just like in Swift, you can also use `else if` statements:
```

For singular it would have to include the word `an` but I think plural fits better here anyway. (same below)

4.0/docs/leaf/overview.md
```diff
@@ -101,7 +101,7 @@ If you want to use another tag as part of your condition, you should omit the `#
 #endif
 ```
 
-You can also use `#elseif` statement.s
+You can also use `#elseif` statement:
```
0xTim, line 104: Nit - spelling

```suggestion
You can also use `#elseif` statements:
```

0xTim (approved):
Thanks! LGTM

PR #583, opened 2021-01-12
Correct the base64 encoding of user credential.

Hailong:
Correct the base64 encoding of user credential `test@vapor.codes:secret42`


0xTim (approved):
Thanks!

PR #585, opened 2021-01-13
Fixing Vapor target dependency for SPM

leogdion:
Fixed Vapor target dependency for SPM. Replaced `Fluent` for `Vapor`.


0xTim (approved):
Thanks!

PR #586, opened 2021-01-16
Fix the table name in foreign key sample code.

Hailong:
The correct table name should be `stars` according to the context.


0xTim (approved):
Thanks!

PR #589, opened 2021-01-25
Add Linux Development hint in hello-world.md

NicFontana:
The development experience in Linux, for people that use editors like VSCode with the [SDE extension](https://marketplace.visualstudio.com/items?itemName=vknabel.vscode-swift-development-environment), is terrible without the `LinuxMain.swift` file. 
That's because the Language Server Protocol doesn't work correctly and the result is: every file becomes red, `no such module` error anytime there is an `import Something` statement, and no features like code completion, quick help (hover), jump to definition, ecc...

This section in the doc can help newbies that want to develop their projects under Linux with ease.

@0xTim

### DIFF
--- a/4.0/docs/hello-world.md
+++ b/4.0/docs/hello-world.md
@@ -27,7 +27,7 @@ open Package.swift
 
 ## Linux Development
 
-If you are working on Linux, consider to add an empty file called `LinuxMain.swift` under the `Sources/Tests/` folder. This allows the [Language Server Protocol](https://github.com/apple/sourcekit-lsp) included in your Swift toolchain to works properly, and if you are working on an editor that supports it, you can have a better development experience.
+If you are working on Linux, consider adding an empty file called `LinuxMain.swift` under the `Sources/Tests/` folder. This allows the [Language Server Protocol](https://github.com/apple/sourcekit-lsp) included in your Swift toolchain to works properly, so if you are working on an editor that supports it, you can have a better development experience.
 
 ## Xcode Dependencies
 
--- a/4.0/docs/hello-world.md
+++ b/4.0/docs/hello-world.md
@@ -25,6 +25,10 @@ cd hello
 open Package.swift
 ```
 
+## Linux Development
+
+If you are working on Linux, consider to add an empty file called `LinuxMain.swift` under the `Sources/Tests/` folder. This allows the [Language Server Protocol](https://github.com/apple/sourcekit-lsp) included in your Swift toolchain to works properly, and if you are working on an editor that supports it, you can have a better development experience.
+
 ## Xcode Dependencies
 
 You should now have Xcode open. It will automatically begin downloading Swift Package Manager dependencies.


0xTim:
@NicFontana how do you pass build arguments to the LSP? There must be a way to pass compiler flags or other build arguments to SourceKit LSP - that's a better route than adding an empty LinuxMain.swift that's now considered deprecated

NicFontana:
@0xTim here is the LSP usage:
```shell
OVERVIEW: Language Server Protocol implementation for Swift and C-based languages

USAGE: sourcekit-lsp [options]

OPTIONS:
  --build-path            Specify build/cache directory
  --completion-max-results
                          When server-side filtering is enabled, the maximum number of results to return
  --completion-server-side-filtering
                          Whether to enable server-side filtering in code-completion
  --configuration, -c     Build with configuration (debug|release) [default: debug]
  --log-level             Set the logging level (debug|info|warning|error) [default: info]
  -Xcc                    Pass flag through to all C compiler invocations
  -Xclangd                Pass options to clangd command-line
  -Xcxx                   Pass flag through to all C++ compiler invocations
  -Xlinker                Pass flag through to all linker invocations
  -Xswiftc                Pass flag through to all Swift compiler invocations
  -index-db-path          Override index-database-path from the build system
  -index-store-path       Override index-store-path from the build system
  --help                  Display available options
```

I can pass flags to the Swift compiler, but it doesn't know anything about the `--enable-test-discovery` SPM flag. 
Do you have any suggestions?

fouadhatem:
Couldn't we just add an extension to SourceKit to accommodate missing flags?

0xTim:
@NicFontana does just passing `--enable-test-discovery` work?

NicFontana:
@0xTim unfortunately no.

`--enable-test-discovery` is a `swift build`/`swift test` flag, It doesn't work if I pass it to `sourcekit-lsp`.

With `sourcekit-lsp` I can pass flags to all Swift compiler invocations using `-Xswiftc`, but since `--enable-test-discovery` is not a Swift compiler (`swift`) flag, it doesn't work either 😞

siemensikkema:
@NicFontana this sounds like a missing feature of sourcekit-lsp to me. Maybe something like an `Xswiftb` option where you can pass `swift build` arguments (but then what about `swift test`?) or even a "top level" `--enable-test-discovery` flag could be an option. Maybe that should even be the default.

I'm not sure what the best place to raise such a feature request would be (other than a PR to sourcekit-lsp itself) but maybe you can try Swift Forums if you're up for it. That would help even more people than having a note in Vapor's README!

fouadhatem:
@NicFontana the best path forward for this would be to raise an issue with `sourcekit-lsp`. It's still early days for them and they continuously add new features. I suggest moving this issue over to them, as there's not much Vapor can do about it.

NicFontana:
@siemensikkema no problem for the Swift Forum request, I can do it.

The note in Vapor's README could be a temporary solution until they figure out what to do with `sourcekit-lsp`. What do you think?

siemensikkema:
@fouadhatem I was about to suggest the same thing but issues are disabled on the repository (like on all Apple repositories?) which is why I suggested Swift Forums.

siemensikkema:
@NicFontana the thing with temporary solutions is that they tend to stick around for longer than needed 😅

fouadhatem:
@siemensikkema oh! I wasn't aware of that 😇

NicFontana:
@siemensikkema I see.

In my humble opinion, not having a useful information today, is worse than having a no longer needed information tomorrow.

0xTim:
LinuxMain.swift is deprecated and we shouldn't be suggesting deprecated stuff in the docs. The error message when you try to compile provides the steps needed to fix it, which I think is enough.

The underlying bug is an issue with SourceKit-LSP, which hopefully they can fix. This also becomes a non-issue in hopefully a few weeks when 5.4 is released

PR #591, opened 2021-03-04
Add optional child documentation

fananek:
Add optional child documentation

### DIFF
--- a/4.0/docs/fluent/relations.md
+++ b/4.0/docs/fluent/relations.md
@@ -69,7 +69,7 @@ A new model can be added to this relation using the `create` method.
 ```swift
 // Example of adding a new model to a relation.
 let jane = Governor(name: "Jane Doe")
-mars.$governor.create(john, on: database)
+mars.$governor.create(Jane, on: database)
 ```
 
 This will set the parent id on the child model automatically.
@@ -89,7 +89,7 @@ database.schema(Governor.schema)
 ```
 !!! warning
     Omitting the unique constraint on the parent ID field from the client's schema can lead to unpredictable results.
-    If there is no uniqueness constraint, the child table may wind up containing more than one child row for any given parent; in this case, an @OptionalChild property will still only be able to access one child at a time, with no way of controlling which child is loaded. If you may need to store multiple child rows for any given parent, use @Children instead.
+    If there is no uniqueness constraint, the child table may wind up containing more than one child row for any given parent; in this case, an `@OptionalChild` property will still only be able to access one child at a time, with no way of controlling which child is loaded. If you may need to store multiple child rows for any given parent, use `@Children` instead.
 
 ## Children
 
--- a/4.0/docs/fluent/relations.md
+++ b/4.0/docs/fluent/relations.md
@@ -89,7 +89,7 @@ database.schema(Governor.schema)
 ```
 !!! warning
     Omitting the unique constraint on the parent ID field from the client's schema can lead to unpredictable results.
-    It means there can potentially exist multiple children while query result will always return the first one.
+    If there is no uniqueness constraint, the child table may wind up containing more than one child row for any given parent; in this case, an @OptionalChild property will still only be able to access one child at a time, with no way of controlling which child is loaded. If you may need to store multiple child rows for any given parent, use @Children instead.
 
 ## Children
 
--- a/4.0/docs/fluent/relations.md
+++ b/4.0/docs/fluent/relations.md
@@ -69,7 +69,7 @@ A new model can be added to this relation using the `create` method.
 ```swift
 // Example of adding a new model to a relation.
 let jane = Governor(name: "Jane Doe")
-mars.$governor.create(Jane, on: database)
+mars.$governor.create(jane, on: database)
 ```
 
 This will set the parent id on the child model automatically.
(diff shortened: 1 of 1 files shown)


4.0/docs/fluent/relations.md
```diff
@@ -50,6 +50,47 @@ The field definition is similar to `@Parent`'s except that the `.required` const
...
+database.schema(Governor.schema)
+    .id()
+    .field("name", .string, .required)
+    .field("planet_id", .uuid, .required, .references("planets", "id"))
+    // Example of unique constraint
+    .unique(on: "planet_id")
+    .create()
+```
+!!! warning
+    Ommiting child's unique constraint on parent id field can lead to unpredictable results.
```
gwynne, line 91:
```suggestion
    Omitting the unique constraint on the parent ID field from the client's schema can lead to unpredictable results.
```

4.0/docs/fluent/relations.md
```diff
@@ -50,6 +50,47 @@ The field definition is similar to `@Parent`'s except that the `.required` const
...
+// Example of adding a new model to a relation.
+let jane = Governor(name: "Jane Doe")
+mars.$governor.create(john, on: database)
+```
+
+This will set the parent id on the child model automatically.
+
+Since this relation does not store any values, no database schema entry is required for the root model.
+
+Child uniqueness should be enforced by `.unique` constraint in child database schema.
```
gwynne, line 79:
```suggestion
The one-to-one nature of the relation should be enforced in the child model's schema using a `.unique` constraint on the column referencing the parent model.
```

4.0/docs/fluent/relations.md
```diff
@@ -50,6 +50,47 @@ The field definition is similar to `@Parent`'s except that the `.required` const
...
+    .id()
+    .field("name", .string, .required)
+    .field("planet_id", .uuid, .required, .references("planets", "id"))
+    // Example of unique constraint
+    .unique(on: "planet_id")
+    .create()
+```
+!!! warning
+    Ommiting child's unique constraint on parent id field can lead to unpredictable results.
+    It means there can potentially exist multiple children while query result will always return the first one.
```
gwynne, line 92: If there is no uniqueness constraint, the child table may wind up containing more than one child row for any given parent; in this case, an `@OptionalChild` property will still only be able to access one child at a time, with no way of controlling which child is loaded. If you may need to store multiple child rows for any given parent, use `@Children` instead.

gwynne (approved):
Looks good, just a few grammatical nitpicks!

4.0/docs/fluent/relations.md
fananek: Docs updated as suggested.

fananek:
@gwynne Thanks for suggestions. PR should be ready for merge now. 👍

4.0/docs/fluent/relations.md
```diff
@@ -50,6 +50,47 @@ The field definition is similar to `@Parent`'s except that the `.required` const
...
+```
+
+The `for` parameter accepts a key path to a `@Parent` or `@OptionalParent` relation referencing the root model.
+
+A new model can be added to this relation using the `create` method.
+
+```swift
+// Example of adding a new model to a relation.
+let jane = Governor(name: "Jane Doe")
+mars.$governor.create(john, on: database)
```
0xTim, line 72: Nit, this is wrong
```suggestion
mars.$governor.create(Jane, on: database)
```

4.0/docs/fluent/relations.md
```diff
@@ -50,6 +50,47 @@ The field definition is similar to `@Parent`'s except that the `.required` const
...
+    .id()
+    .field("name", .string, .required)
+    .field("planet_id", .uuid, .required, .references("planets", "id"))
+    // Example of unique constraint
+    .unique(on: "planet_id")
+    .create()
+```
+!!! warning
+    Omitting the unique constraint on the parent ID field from the client's schema can lead to unpredictable results.
+    If there is no uniqueness constraint, the child table may wind up containing more than one child row for any given parent; in this case, an @OptionalChild property will still only be able to access one child at a time, with no way of controlling which child is loaded. If you may need to store multiple child rows for any given parent, use @Children instead.
```
0xTim, line 92: Nit - wrap the property wrappers in code blocks

```suggestion
    If there is no uniqueness constraint, the child table may wind up containing more than one child row for any given parent; in this case, an `@OptionalChild` property will still only be able to access one child at a time, with no way of controlling which child is loaded. If you may need to store multiple child rows for any given parent, use `@Children` instead.
```

0xTim (changes requested):
Last few nits

4.0/docs/fluent/relations.md
```diff
@@ -50,6 +50,47 @@ The field definition is similar to `@Parent`'s except that the `.required` const
...
+```
+
+The `for` parameter accepts a key path to a `@Parent` or `@OptionalParent` relation referencing the root model.
+
+A new model can be added to this relation using the `create` method.
+
+```swift
+// Example of adding a new model to a relation.
+let jane = Governor(name: "Jane Doe")
+mars.$governor.create(Jane, on: database)
```
0xTim, line 72: This should be lowercase, looks like an autocorrect on my part. I can fix this 
```suggestion
mars.$governor.create(jane, on: database)
```

0xTim (approved):
LGTM!

PR #593, opened 2021-03-07
Fix links in relations chapter

fananek:
Fix links in relations chapter.
- add reference to one-to-one relation
- add missing space in Optional Child header
- fix `load` link broken by #580 

Fix links in relations chapter.


0xTim (approved):
Thanks!

PR #595, opened 2021-03-18
Replace the occurrences of "router.get" in sample code with "app.get" for 4.0 docs.

Hailong:
I just noticed that there are some places in 4.0 document are using `router.get` in sample code, while that seems to be a legacy pattern in earlier versions. So this change is to replace them with `app.get`.


0xTim (approved):
Thanks!

PR #596, opened 2021-03-18
Add step to generate LinuxMain.swift for Heroku deployment.

Hailong:
Heroku will report the following error in build with the default Vapor scaffolding, we'd better add this step in the doc to overcome this issue.

```
error: missing LinuxMain.swift file in the Tests directory
 !     Push rejected, failed to compile Swift app.
 !     Push failed
```

### DIFF
--- a/4.0/docs/deploy/heroku.md
+++ b/4.0/docs/deploy/heroku.md
@@ -134,6 +134,16 @@ echo "web: Run serve --env production" \
   "--hostname 0.0.0.0 --port \$PORT" > Procfile
 ```
 
+### LinuxMain.swift
+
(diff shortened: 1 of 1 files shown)


0xTim:
This has come up several times before. The Core Team (and swift itself) considers LinuxMain deprecated and legacy. The real solution is to use test discovery and instructions for Heroku are [in the build pack docs](https://github.com/vapor-community/heroku-buildpack#test-discovery) 

This all goes away in (hopefully) a couple of weeks when Swift 5.4 is released as test discovery becomes enabled by default.

Hailong:
@0xTim Haha, thanks for letting me know. I'm going to check out the build pack docs.

PR #600, opened 2021-05-12
Fixed branch name in GH Action

MrWoWander:
The project did not deploy because the branch name was incorrect


0xTim (approved):
Good catch, thanks!

PR #608, opened 2021-06-23
Fix example response in Content > Query

commscheck:
The response now matches its description and the example code that generates it.


commscheck:
Realised my mistake, it's referring to the request. Please disregard 😅

PR #609, opened 2021-07-05
Update nginx.md

armanism24:
Added tutorials of installing Nginx on ubuntu 21.04


0xTim:
Swift is not available (officially) on 21.04 therefore Vapor does not support 21.04

PR #613, opened 2021-07-30
Update official MongoDB client reference to mention Vapor integration

kmahar:
Now that mongodb-vapor exists, it would be nice to include a mention of that here too since it will provide a better experience than using the driver directly.


0xTim (approved):
Thanks!

PR #614, opened 2021-08-03
Clarified environment setting with supervisor.

jaanus:
I needed to set multiple environment values with supervisor configuration. Found the part in supervisor docs that explains it, bringing it over to the Vapor guide.


0xTim (approved):
Thanks!

PR #627, opened 2021-11-14
Update to reflect currently implemented database transaction api

dtaylor1701:
`FluentKit` offers a closure-based method for using a database transaction with async functions. This updates the example to reflect that.


0xTim (approved):
Thanks!

PR #630, opened 2021-12-06
Fix note and warning admonitions rendering

kwridan:
- A minor fix for the [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) used for the leaf docs
- Replaced inline **notes** with equivalent admonitions

Test Plan:

- Browse the rendered leaf docs

```sh
pip3 install -r requirements.txt
cd 4.0
mkdocs serve
```

Before:

before

After:

after


0xTim (approved):
Thanks!

PR #631, opened 2021-12-09
Fix function signature from beforeDecode to beforeEncode

wnagrodzki:
Found and fixed a typo in function name.

Fix function signature from beforeDecode to beforeEncode


0xTim (approved):
Thanks!

PR #634, opened 2022-01-15
refactor: remove using deprecated method in Heroku deploy

piersebdon:
`.forClient()` has been deprecated.

This change uses the recommended renamed method in the deprecated message  -> `Use 'makeClientConfiguration()' instead`


0xTim (approved):
Thanks!

PR #642, opened 2022-02-27
Add information about creating tables with parent-children relations

jboulter11:
I've added some information about creating tables with parent-children relationships.  Specifically pointing out a common mistake about creating tables at the same time which contain a reference, and emphasizing that parents don't have a column for their children.

When I was creating my first tables, I ran into this error, which, after I resolved it, prompted this PR:
```
[ WARNING ] server: relation "table_name" does not exist (RangeVarGetRelidExtended)
Swift/ErrorType.swift:200: Fatal error: Error raised at top level: server: relation "table_name" does not exist (RangeVarGetRelidExtended)
```

and then later this:
```
server: null value in column "child_ids" of relation "parent" violates not-null constraint
```

I'm new to Vapor/Fluent so let me know if this info is not good guidance/not correct.  Let me know what you'd like changed.

### DIFF
--- a/4.0/docs/fluent/relations.md
+++ b/4.0/docs/fluent/relations.md
@@ -126,7 +126,21 @@ try await sun.$planets.create(earth, on: database)
 
 This will set the parent id on the child model automatically.
 
-Since this relation does not store any values, no database schema entry is required. 
+!!! warning Since this relation does not store any values, no database schema entry is required for the parent table.  Creating one will likely lead to errors like `server: null value in column "planet_ids" of relation "star" violates not-null constraint`.
+
+
+If a table or collection does not yet exist, and you attempt to reference it, an error will be thrown.  You must create the parent table first, then the children table.
+
+```swift
+// Create the tables
+try await db.schema("stars")
+    .id()
+    .create()
+try await db.schema("planets")
+    .id()
+    .field("star_id", .uuid, .required, .references("stars", "id"))
+    .create()
+```
 
 ## Siblings
 
@@ -342,7 +356,3 @@ planet.$star.value = star
 ```
 
 This will attach the related model to the parent as if it was eager loaded or lazy loaded without an extra database query.
-
-## Migrations
-
-Sometimes setting up relations may require extra work during a migration.  See [schema builder](schema.md) for more info.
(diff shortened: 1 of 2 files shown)


4.0/docs/fluent/schema.md
```diff
@@ -186,7 +186,7 @@ To add a foreign key constraint to a field, use `.references`.
 .field("star_id", .uuid, .required, .references("stars", "id"))
 ```
 
-The above constraint requires that all values in the "star_id" field must match one of the values in Star's "id" field.
+The above constraint requires that all values in the "star_id" field must match one of the values in Star's "id" field, and that the "stars" table already exist.
```
0xTim, line 189:
```suggestion
The above constraint requires that all values in the `"star_id"` field must match one of the values in Star's `"id"` field, and that the `"stars"` table already exist.
```

4.0/docs/fluent/relations.md
```diff
@@ -126,7 +126,20 @@ try await sun.$planets.create(earth, on: database)
...
+```swift
+// Create the tables
+try await db.schema("stars")
+    .id()
+    .create()
+try await db.schema("planets")
+    .id()
+    .field("star_id", .uuid, .required, .references("stars", "id"))
+    .create()
+```
```
0xTim, line 142: This is schema stuff so should live elsewhere, I suggest we remove it

4.0/docs/fluent/relations.md
```diff
@@ -126,7 +126,20 @@ try await sun.$planets.create(earth, on: database)
 
 This will set the parent id on the child model automatically.
 
-Since this relation does not store any values, no database schema entry is required. 
+!!! warning Since this relation does not store any values, no database schema entry is required for the parent table.  Creating one will likely lead to errors like `server: null value in column "planet_ids" of relation "star" violates not-null constraint`.
```
0xTim, line 129: I think the changes in this file can be reverted. I think they'll confuse more than help

0xTim (changes requested):
@jboulter11 sorry for taking so long to get to this! I've got some comments, what do you think

dannflor:
Closing as stale.

PR #646, opened 2022-04-20
Update Heroku Postgres section for mandatory TLS

vzsg:
Heroku does not differentiate between free/hobby and standard plans anymore, unverified TLS must be used for all plans – making the first snippet obsolete and harmful towards new users.


0xTim (approved):
🙌

PR #662, opened 2022-05-11
Add German Translations

0xTim:
Here's the PR. I think the only outstanding thing to fix is updating the index page then we're good to go


mattesmohr:
Yup, you are right. Will do.

mattesmohr (approved):
Looks good to me.

PR #668, opened 2022-06-16
Add German translation

mattesmohr:
This PR is a draft for the German translation of the documentation.


mattesmohr:
I will split up the previous branch for the German translation, since it's a lot of commits behind and do not want to make a mess.

PR #671, opened 2022-06-27
Fix custom-tags to make it clear where the implementation is

dannflor:
The implementation to `render` is just hanging out there in the code block. Enclosed it as a function body.


0xTim (changes requested):
Thanks for this! One small nit around indentation then we're good

0xTim (approved):
Thanks!

Issue #665, opened 2022-05-16
Autocreate translation issues


0xTim:
We should create a GH action that runs whenever a PR is merged to create an issue for translating the documentation issues in that PR. The issue body should contain a link to the PR so we can link them up. This ensures that the translations don't get out of sync.

Even better if we can add a later to the PR to say `no-doc-update` required so it doesn't create issues with that label, for instance if we're changing the MKDocs layout, CI or a translation itself.

mattesmohr:
That's a good idea. Pity I don't nothing about GH Actions. 

Till then, I found the CC (like the last time @0xTim ) helpful.

PR #674, opened 2022-07-03
Create gh action for creating translation issues

jroya1:
### Create a github action action as described in https://github.com/vapor/docs/issues/665 

The only issue with I'm having with it is that it builds a container for the issue creating action before we check if we want to create an issue. So essentially we are using 10-20s each run for a thing that may be not used.

 Please let me know what you want to see in the issue title and body. Right now it's "Need translation for *link to PR*"

And the `no-doc-update` where should we check for it? The current implementation looks for in the PR body

### DIFF
--- a/.github/workflows/translation-issue-template.md
+++ b/.github/workflows/translation-issue-template.md
@@ -1,0 +1,13 @@
+---
+title: Translation needed for {{'#'}}{{env.PR_NUMBER}}
+---
+
+The docs have been updated in PR {{'#'}}{{env.PR_NUMBER}}. The translations should be updated if required.
+
+Languages:
+- [ ] English
+- [ ] Chinese
+- [ ] German
+- [ ] Dutch
+
+Assigned to @vapor/translations - please submit a PR with the relevant updates and check the box once merged. Please ensure you tag your PR with the `translation-update` so it doesn't create a new issue!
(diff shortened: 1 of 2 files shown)


0xTim:
Ok well this is awesome to start with! So what I'd like to ideally do is check for a `no-translation-needed` label as to whether or not to create the issue. And if possible assign the `@vapor/translations` team to the issue so it's picked up by all the language owners. Finally if we could add a checklist for the different languages in the body that would be the icing on the cake

jroya1:
Sounds doable to me! I will make the changes as soon as I can. Just to clarify, the checklist should be just static text for each issue?

0xTim:
Yeah. Could you also add a check for a `translation-update` label as well as `no-translation-needed`? The body can look like something along the lines of:

The docs have been updated in PR #671. The translations should be updated if required.

Languages:

- [ ] English
- [ ] Chinese
- [ ] German
- [ ] Dutch

Assigned to @vapor/translations - please submit a PR with the relevant updates and check the box once merged. Please ensure you tag your PR with the `translation-update` so it doesn't create a new issue!

0xTim:
This is great! The only thing we need to change is that it looks like the assignees field does not accept teams 😒 https://github.com/dear-github/dear-github/issues/170

So I think for now we can just remove that part and hopefully tagging the team will be enough. We can also improve it if it doesn't seem to work

Issue #678, opened 2022-07-14
Check links in CI


0xTim:
Links change when we're updating docs etc, we should make sure we don't break anything in CI.
See https://pypi.org/project/mkdocs-linkcheck/ and https://stackoverflow.com/questions/67924428/how-to-check-dead-links-in-a-gitlab-ci-job-which-builds-a-static-site-with-mkdoc

kazogihara:
@0xTim 
This issue has been done so this should be closed.

0xTim:
Closed via #679

0xTim:
@twof still considering this!

PR #679, opened 2022-07-14
added dead link check CI #678 

kazogihara:
Hi! I just added CI and fixed all detected dead links so could you check it?  

1. Added dead link check CI.
  * Using markdown-link-check.
  * CI check all documents links.

2. Fix Dead link
  * Some links had been broken so pass the CI for PR I have fixed them.
  * schema.zh.md is referred to but not exists so I created it.

3. Notification
  * docs\upgrading.md has some diff debris, please clean it up.


docs/fluent/schema.zh.md
```diff
@@ -0,0 +1,392 @@
+# Schema
```
0xTim, line 1: We can remove this file. The translation plugin will automatically fall back to the default page if no translation is available

0xTim (commented):
Thanks for fixing all these links! I'm curious why you chose an NPM project for the link checking instead of using the [MKDocs package](https://pypi.org/project/mkdocs-linkcheck/) (which is what builds our docs anyway)? My main concern is that if anyone wants to run it locally they need to set up NPM and install everything instead of using the existing ecosystem we already use for the docs

kazogihara:
@0xTim 
> Thanks for fixing all these links! I'm curious why you chose an NPM project for the link checking instead of using the [MKDocs package](https://pypi.org/project/mkdocs-linkcheck/) (which is what builds our docs anyway)? My main concern is that if anyone wants to run it locally they need to set up NPM and install everything instead of using the existing ecosystem we already use for the docs

Thanks for the review!
With mkdocs-linkcheck we have to specify internal links like '../crypto#bcrypt'(the original is crypto.md#bcrypt which is appeared at security/password.md) even if crypto.md exists in current directory.
It still works but it is a bit odd. Because when I click links with vscode, file not found.
mkdocs-linkcheck is nice but also markdown-link-check is good with docker.
Which do u like?

mattesmohr:
Yes, it seems like the ../ prefix works in visual studio code only for the root folder. So in this case visual studio code accepts '../security/crypto.md' or just 'crypto.md'.

@kazogihara But where did you find the information about specification for the link? If I run the check it says 'broken links' anyway. No matter what or how I specify the link.

@0xTim maybe take into consideration that repository for the **mkdocs-linkcheck** doesn't exit anymore. So it doesn't get maintained.

kazogihara:
@mattesmohr 
Thanks for joining review!

> @kazogihara But where did you find the information about specification for the link? If I run the check it says 'broken links' anyway. No matter what or how I specify the link.

I found out by similar [issue](https://github.com/byrnereese/linkchecker-mkdocs/issues/3)

Original
test1


Fixed
test2

> @0xTim maybe take into consideration that the repository for the mkdocs-linkcheck doesn't exit anymore. So it doesn't get maintained.

The [repository](https://github.com/byrnereese/linkchecker-mkdocs) exists. However, updates stopped a year ago :P

>

docs/fluent/schema.zh.md
kazogihara: OK! I will remove it

mattesmohr:
@kazogihara. I see. But I had to remove the file extension too. Now it sees the link as correct. And it looks like mkdocs manage it anyway. The redirect works too.

You are right! My bad. The project link on the https://pypi.org led me to 404. xD

docs/fluent/schema.zh.md
kazogihara: @0xTim 
Unfortunately, it looks like if the file doesn't exist, it's a dead link.

localizationtest


localizationnotfound

0xTim: Yeah I think the link is just wrong and wasn't updated when we did some reorganising. It should be https://docs.vapor.codes/zh/fluent/schema/ - that works

0xTim:
Ok so regarding the NPM link checker - if the MKDocs one is not getting updates let's use the NPM one

docs/fluent/schema.zh.md
kazogihara: I see! I can confirm that it does indeed work in my environment!

kazogihara:
@0xTim 
I fixed several points so if everything ok, please merge it?
* Remove schema.zh.md and fix local links to transmit correctly.
* Add markdown-link-check config for testing locally.
* Add description how to test locally.

docs/fluent/migration.zh.md
```diff
@@ -93,6 +93,10 @@ try await app.autoMigrate()
 
 ## 下一步
 
+<<<<<<< HEAD
+请查看[schema builder](schema.md) 和 [query builder](query.md) 指南，以了解更多迁移相关的信息。
+=======
 请查看[schema builder](schema.md) 和 [query builder](query.zh.md) 指南，以了解更多迁移相关的信息。
+>>>>>>> main
```
0xTim, line 100: There's a left over merge conflict here

0xTim (changes requested):
Thanks for this! Just one last issue to solve

0xTim (approved):
Thanks!

PR #693, opened 2022-07-29
add advanced services translation

redsun:
@Jinxiansen review~


Jinxiansen (approved):
Thanks for your contribution!

Jinxiansen:
image

@0xTim  Hi Tim, do you know about this error? when the action event is compiled?

0xTim:
@Jinxiansen It means the link in the docs is not returning a successful status code. It's probably a flaky check so if you rerun it then it should be fine

PR #700, opened 2022-08-11
Added closure documentation for redirect

iamtuxmeister:
This describes how to use the closure argument to the RedirectMiddleware to generate the string to redirect to instead of a static argument.
Also catches a pitfall for RedirectMiddleware by suggesting the use of User.SessionAutheticator() to manage logged in users.

### DIFF
--- a/docs/security/authentication.md
+++ b/docs/security/authentication.md
@@ -756,7 +756,21 @@ When protecting routes for an API, you traditionally return an HTTP response wit
 let protectedRoutes = app.grouped(User.redirectMiddleware(path: "/login?loginRequired=true"))
 ```
 
+The `RedirectMiddleware` object also supports passing a closure that returns a `String` during creation for advanced url handling. For instance, uncluding the path redirected from as query paramater to the redirect target for state management.
+
+```swift
+let redirectMiddleware = User.redirectMiddleware { req -> String in
+  return "/login?authRequired=true&next=\(req.url.path)"
+}
+```
+
 This works similar to the `GuardMiddleware`. Any requests to routes registered to `protectedRoutes` that aren't authenticated will be redirected to the path provided. This allows you to tell your users to log in, rather than just providing a **401 Unauthorized**.
+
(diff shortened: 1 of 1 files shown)


docs/security/authentication.md
```diff
@@ -756,8 +756,22 @@ When protecting routes for an API, you traditionally return an HTTP response wit
 let protectedRoutes = app.grouped(User.redirectMiddleware(path: "/login?loginRequired=true"))
 ```
 
+The `RedirectMiddleware` object also supports passing a closure that returns a `String` during creation for advanced url handling. For instance, uncluding the path redirected from as query paramater to the redirect target for state management.
```
0xTim, line 759: Spelling and clarification
```suggestion
The `RedirectMiddleware` object also supports passing a closure that returns the redirect path as a `String` during creation for advanced url handling. For instance, including the path redirected from as query parameter to the redirect target for state management.
```

docs/security/authentication.md
```diff
@@ -756,8 +756,22 @@ When protecting routes for an API, you traditionally return an HTTP response wit
...
+
+```swift
+let redirectMiddleware = User.redirectMiddleware { req -> String in
+  return "/login?authRequired=true&next=\(req.url.path)"
+}
+```
+
 This works similar to the `GuardMiddleware`. Any requests to routes registered to `protectedRoutes` that aren't authenticated will be redirected to the path provided. This allows you to tell your users to log in, rather than just providing a **401 Unauthorized**.
 
+Be sure to include a Session Authenticator before the `RedirectMiddleware` to prevent subsequent requests to the protected routes from asking for a password.
```
0xTim, line 769: Let's clarify why
```suggestion
Be sure to include a Session Authenticator before the `RedirectMiddleware` to ensure the authenticated user is loaded before running through the `RedirectMiddleware`.
```

0xTim (changes requested):
Thanks! Couple of minor tweaks

0xTim (approved):
Thanks!

PR #705, opened 2022-08-15
Fixed API call in Queues docs

Flerex:


Flerex:
My bad, I thought there would be a common API.

0xTim:
@Flerex in an ideal world there would be but not currently! This PR is still valuable though, we should just split it up to say "if calling from an `Request` context..... If calling from an `Application` context... etc

PR #707, opened 2022-08-15
Replace old RediStack API links

BennyDeBock:
RediStack documentation has moved from docs.redistack.info to swiftpackageindex. Hence the link referencing the first site were dead links. These have been replaced with the updated links


0xTim (approved):
Thanks!

PR #709, opened 2022-08-16
Application access clarification on Queues

franciscobretal:
I added some clarification on how to dispatch a job from within the `Application` object. Let me know wether this is okay or it actually needs some tweaking.


0xTim (approved):
Thanks!

jdmcd (approved):

PR #741, opened 2022-09-25
Fix type declaration in AsyncModelMiddleware example

neilhamilton:
The example for defining a model middleware has an incorrect type declaration for the `next` value in the middleware method `create` when using async/await.

Copying and pasting this will cause errors and may take a user some time to figure out what's wrong.

The type for `next` in the async/await middleware method `create` has been updated to `AnyAsyncModelResponder` so it conforms to `AsyncModelMiddleware` protocol.


0xTim (approved):
Thanks!

PR #744, opened 2022-09-28
Use of regular expressions for MongoDB

julianiaccopucci:
The API allows to create a Document with a regular expression to query for items, but there's no documentation about it.

Ideally the key "name" is created using FieldKey.makeMongoKey(), but the extension is internal, so we must hardcode it for now.

### DIFF
--- a/docs/fluent/advanced.md
+++ b/docs/fluent/advanced.md
@@ -161,6 +161,23 @@ This is accepted in MongoDB to access nested values.
 ```swift
 Something.query(on: db).filter("document.key", .equal, 5).first()
 ```
+### Use of regular expressions
+
+You can query MongoDB using the `.custom()` case, and passing a regular expression. [MongoDB](https://www.mongodb.com/docs/manual/reference/operator/query/regex/) accepts Perl compatible regular expressions. 
+
+For example, you can query for case insensitive characters under the field `name`
+
+```swift
+import FluentMongoDriver
+       
+var queryDocument = Document()
+queryDocument["name"]["$regex"] = "e"
+queryDocument["name"]["$options"] = "i"
+
+let planets = try Planet.query(on: req.db).filter(.custom(nameDocument)).all()
+        
+```
+This will return planets containing 'e' and 'E'. You can also create any other complex RegEx accepted by MongoDB
 
 ### Raw Access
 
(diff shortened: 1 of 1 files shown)


docs/fluent/advanced.md
```diff
@@ -161,6 +161,23 @@ This is accepted in MongoDB to access nested values.
 ```swift
 Something.query(on: db).filter("document.key", .equal, 5).first()
 ```
+### Use of regular expressions
+
+You can query MongoDB using the `.custom()` case, and passing a regular expression. [MongoDB](https://www.mongodb.com/docs/manual/reference/operator/query/regex/) accepts Perl compatible regular expressions. 
+
+For example, you can query for case insensitive characters under the field `name`
```
0xTim, line 168:
```suggestion
For example, you can query for case insensitive characters under the field `name`:
```

docs/fluent/advanced.md
```diff
@@ -161,6 +161,23 @@ This is accepted in MongoDB to access nested values.
...
+import FluentMongoDriver
+       
+var queryDocument = Document()
+queryDocument["name"]["$regex"] = "e"
+queryDocument["name"]["$options"] = "i"
+
+let planets = try Planet.query(on: req.db).filter(.custom(nameDocument)).all()
+        
+```
+This will return planets containing 'e' and 'E'. You can also create any other complex RegEx accepted by MongoDB
```
0xTim, line 180:
```suggestion

This will return planets containing 'e' and 'E'. You can also create any other complex RegEx accepted by MongoDB.
```

docs/fluent/advanced.md
```diff
@@ -161,6 +161,23 @@ This is accepted in MongoDB to access nested values.
...
+
+```swift
+import FluentMongoDriver
+       
+var queryDocument = Document()
+queryDocument["name"]["$regex"] = "e"
+queryDocument["name"]["$options"] = "i"
+
+let planets = try Planet.query(on: req.db).filter(.custom(nameDocument)).all()
+
```
0xTim, line 178:
```suggestion
let planets = try Planet.query(on: req.db).filter(.custom(nameDocument)).all()
```

0xTim (approved):
Thanks!

PR #752, opened 2022-10-31
Update nginx.md to remove duplicate

Umar-M-Haroon:
Fixed where there's a duplicate  proxy_pass_header Server

Remove a duplicate header entry in the nginx page of the docs
 Remove duplicate header in nginx docs


0xTim (changes requested):
Thanks! Could you change it in the translations as well as it's just the code block?

Umar-M-Haroon:
updated!

0xTim (approved):
Thanks!

PR #762, opened 2023-01-24
Make custom Leaf tag examples complete

vzsg:
As reported in [vapor/vapor#2942](https://github.com/vapor/vapor/issues/2942), UnsafeUnescapedLeafTag is going unnoticed.

To bring more attention to it, I:
- moved the tip to the top section in all languages,
- updated the HelloTag to use the protocol (since it was rendering garbage since the LeafKit update)
- and arranged all code samples to a common structure so they are now directly copiable.


0xTim (approved):
🙌

Issue #764, opened 2023-02-03
DigitalOcean deploy docs dangerously outdated


gwynne:
The existing DO deployment documentation (http://docs.vapor.codes/deploy/digital-ocean/) is absurdly out of date and does not work as written.

beastboym:
I would like to work but i'm a first timer.
Could you put me through on the details so I could know what to do?

0xTim:
@beastboym looks like you've been beaten to it unfortunately! However you can see our contributing guide here https://docs.vapor.codes/contributing/contributing/

PR #765, opened 2023-02-04
Update Digital Ocean deploy

Tyler-Waitt:
Hi all. This is my first PR (for any project!). Please let me know if I messed anything up. I saw [this issue](https://github.com/vapor/docs/issues/764) on the Digital Ocean deployment docs being out of date and gave it a go. Thanks!


Tyler-Waitt:
Also, I ran into [this bug](https://github.com/vapor/vapor/issues/2681) when working on this guide on Digital Ocean. Adding a swap following [this guide](https://www.digitalocean.com/community/tutorials/how-to-add-swap-space-on-ubuntu-20-04) fixed it. I don't know if that should go in as a note.

docs/deploy/digital-ocean.md
```diff
@@ -117,59 +119,79 @@ Verify that Swift was installed correctly.
...
 ```
 
-If you used `swift build -c release`, then you need to run:
+### Run
+
+Now that Vapor is setup and we have port 80 open, let's run it. 
+
 ```sh
-sudo .build/release/Run serve -b 157.245.244.228:80
+vapor run serve --hostname 0.0.0.0 --port 80 &
```
0xTim, line 181: Does this work? I'd expect to need sudo privileges to run on port 80 (which it shouldn't have!)

I'm fine with using 8080 and opening up the firewall to that port instead.

We also probably don't want to immediately background the process for the demo

0xTim (changes requested):
Thanks for doing this! I have one question around the port.

We can probably point people to the Vapor docs for installing Swift and the toolbox rather than duplicating it here, but until installing Swift is easier I'm happy to keep it

docs/deploy/digital-ocean.md
Tyler-Waitt: Oh, you're right. It would be better to do all of this on 8080. I think I was on root 😬

docs/deploy/digital-ocean.md
```diff
@@ -167,31 +167,31 @@ cd HelloWorld
...
 ```
 
 ### Run
 
-Now that Vapor is setup and we have port 80 open, let's run it. 
+Now that Vapor is setup and we have an open port, let's run it. 
 
 ```sh
-vapor run serve --hostname 0.0.0.0 --port 80 &
+vapor run serve --hostname 0.0.0.0 --port 8080
```
Tyler-Waitt, line 181: Background process removed too

0xTim (approved):
Great thanks!

PR #774, opened 2023-02-28
Add German translation for async.md

mattesmohr:
I will split up the previous branch for the German translation, since it's a lot of commits behind and do not want to mess things up.

### DIFF
--- a/docs/basics/async.de.md
+++ b/docs/basics/async.de.md
@@ -1,0 +1,435 @@
+# Async
+
+## Async Await
+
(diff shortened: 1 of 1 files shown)


0xTim (approved):
LGTM though I don't speak German 🤣 I'm happy to review unless we can pull someone in to check it quickly, maybe @marius-se ?

mattesmohr:
> LGTM though I don't speak German 🤣 I'm happy to review unless we can pull someone in to check it quickly, maybe @marius-se ?

Haha right. @Suboptimierer could do. But still there a some unfinished text passages, where I couldn't find the right translation for.

We could wave it through if you don't mind.

0xTim:
I'll merge them both and we can always tweak them after if needed

PR #781, opened 2023-03-12
Fix getting started URLs by adding /hello

rph8:
http://127.0.0.1:8080 gave me an error message in the browser:
`{"error":true,"reason":"No template found for index"}`
If you add `/hello` in both cases, everything works.

### DIFF
--- a/docs/getting-started/hello-world.de.md
+++ b/docs/getting-started/hello-world.de.md
@@ -53,6 +53,6 @@ Bei der Erstausführung werden die Paketverweise nachgeladen. Dementsprechend ka
 
 ## Aufruf im Browser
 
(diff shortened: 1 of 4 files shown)


0xTim:
@rph8 What project/template are you using? If you use the Vapor template with the options in the docs you'll end up with [this route](https://github.com/vapor/template-bare/blob/main/Sources/App/routes.swift#L4) which has what you need

rph8:
Yes, you are right. So I tried it without the `-n` and then there will not be a route for http://127.0.0.1:8080 apparently, leading to said error message.
Maybe we can add such a note to the docs?
I think some people might want to try it without `-n` and with such a note they would be less confused.

0xTim:
Did you choose an option for Leaf? And if so did you set a custom working directory?

PR #782, opened 2023-03-13
Update Carbon placement

sayzlim:
Update the ad format and only load the ad when it's visible. Here is how the new format and placement looks like.

image


0xTim (approved):
Thanks!

PR #790, opened 2023-04-14
Update queues.md

gargs:
- adds a warning about trying to scheduled jobs outside of boot up

### DIFF
--- a/docs/advanced/queues.md
+++ b/docs/advanced/queues.md
@@ -309,6 +309,9 @@ If you do not specify a queue the job will be run on the `default` queue. Make s
 
 The Queues package also allows you to schedule jobs to occur at certain points in time.
 
+!!! warning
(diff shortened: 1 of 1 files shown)


docs/advanced/queues.md
```diff
@@ -309,6 +309,9 @@ If you do not specify a queue the job will be run on the `default` queue. Make s
 
 The Queues package also allows you to schedule jobs to occur at certain points in time.
 
+!!! warning
+Scheduled jobs only work when set up before the application boots up, such as in configure.swift. They will not work in route handlers.
```
0xTim, line 313:
```suggestion
    Scheduled jobs only work when set up before the application boots up, such as in `configure.swift`. They will not work in route handlers.
```

0xTim (approved):
Thanks!

PR #798, opened 2023-04-27
Revise latest changes of the German translation

mattesmohr:
This PR revises some of the latest changes of the German translation.


0xTim (approved):
Thanks!

PR #805, opened 2023-04-30
Revise queues.md

mattesmohr:
This PR revises some bits of the queues.md, wich come to my attention while working with the library.

### DIFF
--- a/docs/advanced/queues.md
+++ b/docs/advanced/queues.md
@@ -65,6 +65,8 @@ If you edit the manifest directly inside Xcode, it will automatically pick up th
 The next step is to configure Queues in `configure.swift`. We'll use the Redis library as an example:
 
 ```swift
+import QueuesRedisDriver
+
 try app.queues.use(.redis(url: "redis://127.0.0.1:6379"))
 ```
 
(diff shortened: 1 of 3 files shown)


mattesmohr:
@0xTim Do we update the package declaration example from time to time, or only when it's a major? The package is currently by version 1.0.4.

mattesmohr:
In my opinion, that can be simplified

```swift
//Register jobs
let emailJob = EmailJob()
app.queues.add(emailJob)
```

to 

```swift
//Register jobs
app.queues.add(EmailJob())
```

Or is there a reason to do it the way it is?

mattesmohr:
The concurrency version of the jobeventdelegate is not mentioned

```swift
struct PluginEventDelegate: AsyncJobEventDelegate {
    
    func dispatched(job: JobEventData) async throws {
    }

    func didDequeue(jobId: String) async throws {
    }

    func success(jobId: String) async throws {
    }

    func error(jobId: String, error: Error) async throws {
    }
}
```

0xTim (approved):

0xTim:
@mattesmohr nah there's no need to bump the versions unless the docs refer to something specifically in a later version that could cause compiler issues.

Re splitting out the job, mainly helpful when you want to pass in stuff, but it's personal preference for an empty initialiser so it can be left.

Re the concurrency stuff, we should probably add that - feel free to do it as a separate PR!

PR #809, opened 2023-05-31
Expand on why `shutdown()` needs to be used in unit tests

MultiColourPixel:
This PR is a follow-up to the conversation around this topic in Discord: https://discord.com/channels/431917998102675485/484454139506458634/1113123137240973404


0xTim (approved):
Thanks!

PR #813, opened 2023-06-01
Update hljs and code blocks style

ptoffy:
- This updates the `highlight.js` version to be able to use better code highlighting
- Also updates the code blocks style


0xTim (commented):
I wonder if we can pull all these files in from design.vapor.codes instead?

ptoffy:
> I wonder if we can pull all these files in from design.vapor.codes instead?

We could pull in the highlighting library but the `syntax.css` file is slightly different

0xTim (approved):

PR #815, opened 2023-06-02
Example to populate pivot properties

alemohamad:
Add explanation on how to use the callback closure in attach method to populate properties in pivot model.


docs/fluent/relations.md
```diff
@@ -158,7 +158,7 @@ final class PlanetTag: Model {
 }
```
gwynne, line 158: GitHub doesn't allow me to leave comments directly on lines that weren't modified in the original PR, but given the new example code below, these additional fields should be added to the `PlanetTag` model starting at line 151:

```swift
    @OptionalField(key: "comments")
    var comments: String?
    
    @OptionalEnum(key: "status")
    var status: PlanetTagStatus?

```
And, to be really and truly pedantic, the `PlanetTagStatus` enum should be defined somewhere to; I'd just sneak it in after line 140:
```swift
    enum PlanetTagStatus: String, Codable { case accepted, pending }
```

gwynne (changes requested):
Looks good, but I've recommended some tweaks for grammar and completeness.

alemohamad:
Thanks @gwynne for your feedback! I believe that this adds more context to the readers. The changes are already applied. 👍

gwynne (approved):

0xTim (approved):
Thanks!

PR #835, opened 2023-06-15
Update some outdated information on the Heroku guide

Jeehut:
I just followed this guide and found several pieces of information that were outdated. Most noticably, there are no free tiers on Heroku any more, so I adjusted the text and added the new price for what was available for free before ($5 per dyno & database).

The text also made it seem like heroku has a preferred branch name and that that is "master". But AFAIK heroku doesn't care about the branch name, it just takes what you push. And there was a movement to no longer name the default branch "master" in our community, so I added the alternative name "main" in the text, too.

### DIFF
--- a/docs/deploy/heroku.md
+++ b/docs/deploy/heroku.md
@@ -13,7 +13,7 @@ Make sure that you've installed the heroku cli tool.
 ### HomeBrew
 
 ```bash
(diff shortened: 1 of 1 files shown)


0xTim (approved):
Thanks!

PR #854, opened 2023-07-09
Update Search Index Script for new languages

### LINKED ISSUE #853: Search contains English content from non-English site translations
### Describe the bug

While using search it is easy to end up on a version of the docs site in a different language.

This happens for two reasons:
1. Search results contain content from all languages, regardless of the user's language (or language of the current site).
2. Many documents appear in search results _in English_ in the translated versions of the site.

While I don't think (1) is a good idea, I could see how this may be intended behaviour, as it may make it easier for non-English speaking users who land on the English site (replace for any other languages) to make their way into their desired site language just via search, rather than via the language menu.

I also understand that (2) may be intended behaviour while docs have yet to be translated, as having an English doc (or other original language) may be better than having a _missing_ doc.

However, the product of these two decisions/bugs/factors is that it is possible to use Search in English, and end up on a different language site unintentionally. Anecdotally, this happens to me _a lot_.

### To Reproduce

1. Go to https://docs.vapor.codes/
2. Search for "docker"
3. Observe 5 results titled "Docker Deploys"
4. Click on a result other than the first
5. Observe that the site is no longer in English

### Expected behaviour

- At step (3), I would expect that there is only a single result for the "Docker Deploys" page.
- At step (5), having followed that single hit, I would expect the site to be in English (or the original browsing language).

I haven't managed to reproduce this in a case where the _top hit_ has not been English, however given the anecdotal frequency of hitting this issue I feel there may be cases of it, or at least the UX of the search is such that clicking on subsequent hits is too easy.

### Environment

N/A. I can provide browser details if necessary but I don't think this is a browser compatibility bug.

0xTim:
Adds the new translations to the search index script to ensure the new languages don't appear as currently everything is duplicated until all pages are fully translated or we can find better plugins that only search in the current language

Resolves #853

### DIFF
--- a/fixSearchIndex.swift
+++ b/fixSearchIndex.swift
@@ -37,17 +37,22 @@ let searchIndex = try JSONDecoder().decode(SearchIndex.self, from: indexData)
 var newSearchIndex = searchIndex
 var searchIndexDocs = [SearchIndexDocs]()
 
+let knownLanguages = [
+        "en",
+        "de",
+        "es",
+        "fr",
+        "it",
+        "ko",
+        "nl",
+        "pl",
+        "zh",
+    ].map { "\($0)/" }
+
+
 for doc in newSearchIndex.docs {
-    if !doc.location.starts(with: "en/")
-      && !doc.location.starts(with: "de/")
-      && !doc.location.starts(with: "es/")
-      && !doc.location.starts(with: "fr/") 
-      && !doc.location.starts(with: "it/") 
-      && !doc.location.starts(with: "ko/") 
-      && !doc.location.starts(with: "nl/") 
-      && !doc.location.starts(with: "pl/") 
-      && !doc.location.starts(with: "zk/") {
-          searchIndexDocs.append(doc)
+    if !knownLanguages.contains(where: { doc.location.starts(with: $0) }) {
+        searchIndexDocs.append(doc)
     }
 }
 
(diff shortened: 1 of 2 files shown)


fixSearchIndex.swift
```diff
@@ -39,10 +39,14 @@ var searchIndexDocs = [SearchIndexDocs]()
...
-      && !doc.location.starts(with: "zh/")
       && !doc.location.starts(with: "de/")
+      && !doc.location.starts(with: "es/")
       && !doc.location.starts(with: "fr/") 
-      && !doc.location.starts(with: "nl/") {
+      && !doc.location.starts(with: "it/") 
+      && !doc.location.starts(with: "ko/") 
+      && !doc.location.starts(with: "nl/") 
+      && !doc.location.starts(with: "pl/") 
+      && !doc.location.starts(with: "zk/") {
```
gwynne, line 49: Might I suggest something easier for others to add to if needed:
```swift
    let knownLanguages = [
        "en",
        "de",
        "es",
        "fr",
        "it",
        "ko",
        "nl",
        "pl",
        "zk",
    ].map { "\($0)/" }
    if !knownLanguages.contains(where: { doc.location.starts(with: $0) }) {
```
Also, did you mean to remove `zh/`?

gwynne (changes requested):

gwynne (approved):
👍

PR #867, opened 2023-07-26
Add new templates for vapor/penny-bot#81

MahdiBM:
Adds new leaf templates that can be used with vapor/penny-bot#81.

We should not remove the old templates for now, until Penny has shown some success.


.github/templates/translation_needed.description.leaf
```diff
@@ -0,0 +1,14 @@
+The docs have been updated in PR #(number). The translations should be updated if required.
```
0xTim, line 1: Will the link be auto-created here?

MahdiBM: Yes it'll appear as a proper link on GitHub.

MahdiBM:
Removed the title template as we can just use that in Penny repo.
Moved the description template on level higher to `.github`.

.github/templates/translation_needed.description.leaf
MahdiBM: like `#867`: #867

MahdiBM:
not sure if i should move the template file down to `/workflows`. 
didn't put it there because it's technically not used by GHA workflows anymore.

.github/templates/translation_needed.description.leaf
MahdiBM: ahhhh now i see ... i need to use `in PR ##(number).` instead of `updated in PR #(number).` Considering one `#` is used by leaf syntax. Updated the PR.

0xTim (changes requested):
Can you update the README to ensure people know to update this file when adding new translations? I think we can just point people at this rather than getting them to use both as it will be the way to do it soon enough

MahdiBM:
the build failure is normal. The new README contains a link to the main branch but the file is not there because this PR is not merged yet.

0xTim (approved):

PR #885, opened 2023-08-06
Add default hint for required values in schema

ptoffy:


docs/fluent/schema.md
```diff
@@ -94,6 +94,9 @@ Supported field constraints are listed below.
 |`.references`|Requires that this field's value match a value in the referenced schema. See [foreign key](#foreign-key)|
 |`.identifier`|Denotes the primary key. See [identifier](#identifier)|
 
+!!! tip
+    If you're using a `.required` field, especially if you're updating a schema rather than creating it, you might need to set a default value, this is done via `.sql(.default("default value"))`.
```
MahdiBM, line 98: I think it could be clearer, like 
```suggestion
!!! tip
    You can use `.sql(.default(DEFAULT_VALUE))` to specify a default value for your migrations.
    This is also useful when updating an existing table with a new `.required` field.
```

dannflor: This tip doesn't make it clear *how* to use the default value (putting it as an argument to the `.field` method). You really ought to have a full codeblock with an example. Something like
```swift
try await database.schema(User.schema)
            .field("joined_at", .date, .required, .sql(.default(SQLRaw("NOW()"))))
            .update()
```

ptoffy: That makes sense, but I think we might just need 
`.field("joined_at", .date, .required, .sql(.default(SQLRaw("NOW()"))))`
It's the schema page after all

Although Mahdi's suggestion is not too precise then😆

gwynne: This is also wrong - `NOW()` should be invoked as `SQLFunction("now")`. Also, it's better to recommend a `.datetime` here, rather than a `.date`.

gwynne (changes requested):

ptoffy:
I added the new section but couldn't think of an `SQLRaw` example so feel free to chime in 🙈

gwynne:
> I added the new section but couldn't think of an `SQLRaw` example so feel free to chime in 🙈

That's a good thing - by and large, one shouldn't have to use `SQLRaw`; the other concrete expression types are _ALWAYS_ to be preferred (in fact, there should be a note to that effect in here somewhere).

docs/fluent/schema.md
```diff
@@ -218,6 +218,21 @@ Below is an example using foreign key actions.
     Foreign key actions happen solely in the database, bypassing Fluent. 
     This means things like model middleware and soft-delete may not work correctly.
 
+## SQL
+
+The `.sql` parameter allows you to add arbitrary SQL to your schema. This is useful for adding specific constraints or data types.
+Quite a common use case is defining a default value for a field:
```
gwynne, line 224:
```suggestion
A common use case is defining a default value for a field:
```

gwynne (approved):
One grammar nit, otherwise looks good

MahdiBM (approved):

PR #914, opened 2023-09-27
Update and fix French nav translations

FlorianClaisse:
Added and fixed translation for the navigation bar in French


0xTim (approved):
Merci!

PR #920, opened 2023-10-03
Update routing.md

KalynDavis:
Fixes typo on `Basics` docs section related to 'redirect(to:type:)' being deprecated and renamed to 'redirect(to:redirectType:)'

### DIFF
--- a/docs/basics/routing.md
+++ b/docs/basics/routing.md
@@ -427,7 +427,7 @@ You can also specify the type of redirect, for example to redirect a page perman
 req.redirect(to: "/some/new/path", redirectType: .permanent)
 ```
 
-The different `RedirectType`s are:
+The different `Redirect`s are:
 
 * `.permanent` - returns a **301 Permanent** redirect
 * `.normal` - returns a **303 see other** redirect. This is the default by Vapor and tells the client to follow the redirect with a **GET** request.
--- a/docs/basics/routing.md
+++ b/docs/basics/routing.md
@@ -424,7 +424,7 @@ req.redirect(to: "/some/new/path")
 You can also specify the type of redirect, for example to redirect a page permanently (so that your SEO is updated correctly) use:
 
 ```swift
-req.redirect(to: "/some/new/path", type: .permanent)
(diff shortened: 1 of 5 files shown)


docs/basics/routing.md
```diff
@@ -424,7 +424,7 @@ req.redirect(to: "/some/new/path")
 You can also specify the type of redirect, for example to redirect a page permanently (so that your SEO is updated correctly) use:
 
 ```swift
-req.redirect(to: "/some/new/path", type: .permanent)
+req.redirect(to: "/some/new/path", redirectType: .permanent)
 ```
 
 The different `RedirectType`s are:
```
0xTim, line 430: This line needs to be updated as well
```suggestion
The different `Redirect`s are:
```

0xTim (changes requested):
@KalynDavis Thanks for this! Since this is just code changes could you update the different translations as well?

docs/basics/routing.md
KalynDavis: @0xTim shouldn't this still be `RedirectType's` since it is providing the different static members for `redirectType`?

0xTim: @KalynDavis `RedirectType` is now deprecated and the new type is `Redirect` so we should rename it just to avoid confusion

KalynDavis: @0xTim Oh, My apologies! I was under the impression that this was `type`->`redirectType`. To clarify, it is `redirectType`->`redirect`?

KalynDavis: Just checked the source code. I see the changes. Thanks!

KalynDavis: @0xTim I've updated the translations to properly reflect the changes. Thanks for pointing that out!

Note: I did notice that the German translation doesn't look to have the same formatting as the other translations. Gives me the impression the information might be outdated as well.

mattesmohr: The German translation is fine. 😊. Just different.

0xTim (approved):
Thanks!

PR #942, opened 2023-11-11
Modernize deploy workflow

gwynne:
Switches to OIDC provider, replaces very old S3 and CloudFront actions with modern CLI usage, de-hardcodes region and S3 bucket name, unlocks pin of old version of CloudFormation deploy action, add concurrency safety.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,60 +1,49 @@
 name: Build and deploy the Vapor documentation
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 on:
   push:
     branches:
       - main
+
 jobs:
   deploy:
     name: Build and deploy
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,60 +1,49 @@
...
   push:
     branches:
       - main
+
 jobs:
   deploy:
     name: Build and deploy
     runs-on: ubuntu-latest
+    permissions: { id-token: write, contents: read }
+    env: { AWS_PAGER: '' }
```
0xTim, line 15: What's the pager used for?

0xTim (approved):

.github/workflows/deploy.yml
gwynne: The V2 CLI enables its pager by default even when output isn't to a terminal; it has to be explicitly disabled and the env var is a particularly convenient way to do so across the board.

PR #944, opened 2023-11-16
Update highlightjs

### LINKED ISSUE #637: Leaf Highlight syntax is out of date
Currently built for Leaf 3, we should add support for Leaf 4

https://github.com/highlightjs/highlight.js/blob/main/src/languages/leaf.js needs updating to current spec

dannflor:
Updated highlightjs to 11.9.0 to include new leaf grammar. Also slimmed it down to only include grammars likely to appear in the documentation (html, css, json, etc). Close #637


docs/javascripts/highlight.min.js
gwynne, line 1: @dannflor I would keep the `pgsql`, `bash`, `diff`, `yaml`, and `xml` grammars also.

gwynne (changes requested):

docs/javascripts/highlight.min.js
0xTim: We definitely use `diff` on the docs site

0xTim: To add to this, this is served by a CDN and compressed so a KB or so isn't the end of the world

dannflor: I went through and added all the missing grammars, there's even a `toml` block at some point

gwynne (approved):

PR #947, opened 2023-11-16
Align apns versions and imports

dannflor:
Make sure that `import APNSCore` and `from: "4.0.0"` are present in all versions of `apns.md` in response to https://github.com/vapor/apns/issues/54.


0xTim (approved):
Thanks!

PR #951, opened 2023-11-25
Fix Typo in Fluent Overview

cedricbahirwe:
Fixed a Typo on the Fluent Overview Doc, replacing `seealso` with `see also`


0xTim:
This is actually correct but now deprecated. You can fix the spelling as per https://github.com/squidfunk/mkdocs-material/issues/2700

0xTim:
@cedricbahirwe if you wanted to PR the fix in from the linked issue feel free!

PR #955, opened 2023-12-16
Add missing `else` to guard in docs

wSedlacek:
The code as is has a syntax error.

### DIFF
--- a/docs/basics/controllers.de.md
+++ b/docs/basics/controllers.de.md
@@ -48,7 +48,7 @@ struct TodosController: RouteCollection {
     }
 
     func delete(req: Request) async throws -> HTTPStatus {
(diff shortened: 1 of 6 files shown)


MahdiBM (approved):

gwynne:
Nice catch!! Can you also fix the same issue in the translations of that file? It looks like most of them have the same problem.

gwynne (changes requested):
See comment above

MahdiBM:
Yeah i was going to ask about the translations too 😅 https://github.com/vapor/docs/tree/main/docs/basics

wSedlacek:
@gwynne Got the other translations fixed! Good eye!

0xTim (approved):
Thanks!

PR #956, opened 2023-12-17
Remove the use of `load(on:)` in relations.md

wojexe:
As I've noticed on a [discord conversation](https://discord.com/channels/431917998102675485/684159753189982218/1152776938281250896), the `load(on:)` method is discouraged to be used and should be replaced with `get(on:)`, which defaults to cache if possible.

### DIFF
--- a/docs/fluent/relations.md
+++ b/docs/fluent/relations.md
@@ -331,7 +331,7 @@ The `with` method accepts an optional closure as a second parameter. This closur
 
 ## Lazy Eager Loading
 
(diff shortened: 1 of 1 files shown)


gwynne:
Great! Can you also add mention (and possibly an example) of `get()`'s `reload` parameter? (The full signature is `get(reload:on:)`, with `reload` defaulting to `false`.)

wojexe:
Actually there is a mention of the `reload:` parameter higher up, in the `#get` section. I just thought that `get(on:)` would be sufficient for this use case. But yeah, on the second thought, people might want to have their data reloaded during eager loading, so I'll include it. 😄

PR #965, opened 2024-02-09
Add note about DTO for parent relationship

0xTim:


gwynne:
I would prefer that we document DTOs as recommended for general use with Fluent models, not just specifically when relations are involved. The `Codable` considerations apply to all Fluent properties, just in less obvious fashion.

gwynne (approved):

PR #968, opened 2024-03-29
Update documentation about ws.send(...) return value

jiahan-wu:
## Description

The current documentation incorrectly states that ws.send(...) returns a value which can be awaited. However, the WebSocket.send method does not have a return value.

## Reference

* [vapor/websocket-kit/Sources/WebSocketKit/Concurrency](https://github.com/vapor/websocket-kit/blob/main/Sources/WebSocketKit/Concurrency/WebSocket%2BConcurrency.swift)

### DIFF
--- a/docs/advanced/websockets.md
+++ b/docs/advanced/websockets.md
@@ -58,11 +58,10 @@ promise.futureResult.whenComplete { result in
 }
 ```
 
-If using `async`/`await` you can `await` on the result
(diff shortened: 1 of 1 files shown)


gwynne (approved):

TillTHMAcc:
i tried to push but it saying acces denied, i have the right remote.

with ssh and http the same.

TillTHMAcc:
problem is solved

PR #973, opened 2024-04-23
update postgres connect example to non-deprecated

jaredh159:
current docs show using a deprecated initializer that produces a warning.


0xTim (approved):
Thanks!

PR #980, opened 2024-05-20
fix typo & add missing 'with' argument label

fang-ling:
fix typo `ClientTokenReponse` => `ClientTokenResponse` (réponse to response)

add missing `with` argument label in the initializer of `SessionToken` to fix "Incorrect argument label in call (have 'with:', expected 'user:')".


0xTim (changes requested):
@fang-ling thanks for this! Since this is only code changes do you think you could update the translations as well? Thanks!

fang-ling:
@0xTim thanks for the suggestions. All the translations have been updated.

0xTim (approved):
Thanks!

PR #981, opened 2024-05-25
Fix Error Middleware docs

m1guelpf:
[`ErrorMiddleware` is not the only middleware added by default](https://github.com/vapor/vapor/blob/cc98361cbec22757a89a9717836293f215a0f67d/Sources/Vapor/Middleware/Application%2BMiddleware.swift#L8). This PR updates that assertion and the replacement instructions to re-introduce `RouteLoggingMiddleware`.


0xTim (approved):
Thanks!

PR #990, opened 2024-07-04
fixed missing s as per Python string expression syntax

asnov:
supervisord uses the Python string expression syntax in its configuration files, like this `%(ENV_X)s `
It's described here: http://supervisord.org/configuration.html#environment-variables
In the "Deploy->Supervisor->Configure"  section of the documentation the final "s" were omitted: 
https://docs.vapor.codes/deploy/supervisor/#configure
so copy-pasting this example was leading to wrong file names of the logs.
For example, instead of the expected file name `hello-stdout.log` I had the wrong `hellotdout.log`.


0xTim (approved):
Thanks!

PR #991, opened 2024-07-21
Update Leaf docs

fpseverino:
Update Leaf docs to accept time zone IDs as argument in `#date` tag


fpseverino:
Looks like there is a broken link in the Korean translation

0xTim (approved):

PR #1001, opened 2024-08-21
JWT docs overhaul

ptoffy:
It might make sense for https://github.com/vapor/jwt-kit/issues/186 to be closed before merging this 

Handy review link https://github.com/vapor/docs/blob/jwt-5/docs/security/jwt.md


docs/security/jwt.md
```diff
@@ -121,208 +143,162 @@ TestPayload(
...
-await app.jwt.keys.addHMAC(key: "foo", digestAlgorithm: .sha256, kid: "a")
-// Add HMAC with SHA-256 signer named "b".
-await app.jwt.keys.addHMAC(key: "bar", digestAlgorithm: .sha256, kid: "b")
+await app.jwt.keys.add(pss: key, digestAlgorithm: .sha256)
 ```
 
-When signing JWTs, pass the `kid` parameter for the desired signer.
+## Key Identifier (kid)
+
+When adding a key to the key collection, you can also specify a key identifier (kid). This is a unique identifier for the key that can be used to look up the key in the collection.
```
0xTim, line 282: Thoughts?
```suggestion
When adding a key to the key collection, you should also specify a key identifier (kid). This is a unique identifier for the key that can be used to look up the key in the collection. 
```

ptoffy: I think it makes sense to leave it as optional as
1. Most people probably don't have use than one key
2. I'd say in most cases you don't know the `kid` beforehand so signing/verifying might fail at first which isn't ideal.

If you _should_ specify it, you'll probably know it

0xTim: Ok works for me

0xTim (approved):

PR #1027, opened 2024-12-10
Add Spanish translation for security/passwords.md

alemohamad:
Add Spanish translation for security/passwords.md

### DIFF
--- a/docs/security/passwords.es.md
+++ b/docs/security/passwords.es.md
@@ -59,9 +59,6 @@ let digest = try app.password.hash("vapor")
 ```
 
 ### Async 
-
-Password hashing algorithms are designed to be slow and CPU intensive. Because of this, you may want to avoid blocking the event loop while hashing passwords. Vapor provides an asynchronous password hashing API that dispatches hashing to a background thread pool. To use the asynchronous API, use the `async` property on a password hasher.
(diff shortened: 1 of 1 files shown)


docs/security/passwords.es.md
```diff
@@ -0,0 +1,89 @@
...
+
+La misma API está disponible en `Application` para su uso durante el arranque.
+
+```swift
+let digest = try app.password.hash("vapor")
+```
+
+### Async 
+
+Password hashing algorithms are designed to be slow and CPU intensive. Because of this, you may want to avoid blocking the event loop while hashing passwords. Vapor provides an asynchronous password hashing API that dispatches hashing to a background thread pool. To use the asynchronous API, use the `async` property on a password hasher.
```
TheHandyOwl, line 63: Delete this paragraph

TheHandyOwl (changes requested):

docs/security/passwords.es.md
```diff
@@ -0,0 +1,89 @@
...
+
```
TheHandyOwl, line 64:
```suggestion

```

jacostaf10 (approved):
LGTM, ready to go!

Issue #850, opened 2023-07-05
Hot reloading - state of the project


landsman:
Hello, I tried to search "hot reload" via search on website, in GitHub organisation, issues etc.
What I found is a few questions about that as well.

- https://github.com/vapor/vapor/issues/389
- https://github.com/vapor/vapor/issues/311
- https://github.com/vapor/vapor/issues/212
- https://github.com/vapor/vapor/issues/181

So what is the state? Does it working, or should, is it present? From what I tried I didn't noticed.

Can you please consider to create a new page, at least under Leaf tab and describe the current situation of Hot reload feature It's fine to say that's not supported at the moment and contribution is welcome, or provide a few links to workarounds, unofficial solutions.

I beliave that this is a quite common thing that every new developer looking for and there is no answer.

ptoffy:
You can disable leaf template caching with `app.leaf.cache.isEnabled = false`, but yeah it makes sense for it to be in the docs

0xTim:
The cache is mentioned [here](https://docs.vapor.codes/leaf/getting-started/) but I'm happy to accept a PR to refine it. You still need to reload the page to see the changes. 

There's also a PoC extension for Vapor that includes live previews of templates - https://github.com/HanneVerstraete/vapor-extension

kublaios:
Using https://github.com/cespare/reflex, you can trigger the `swift run` command whenever the files under a certain directory change like so:
```
reflex -r 'Sources\/.*\.swift$' -R '\.build' -s swift run
```

Although I haven't tried it with a large project where the builds take some time.

dector:
> Using https://github.com/cespare/reflex, you can trigger the `swift run`

Sadly, that's still very ugly workaround which even making server "offline" for a sec or two on "hello world" example.

In the roadmap (#181) "Hot Reload" is marked as done. But it's effectively absent. What is the status of this feature? Is further improvement is planned for the future or core team is happy with it as it is?

Thanks!

0xTim:
@dector the issues are probably closed more of an out of scope than fixed. It's out of scope for Vapor itself to implement and something better suited to Swift itself as proper hot loading will require compiler and runtime work

PR #1053, opened 2025-01-15
Refine Leaf cache documentation

alemohamad:
Added an improvement for the cache documentation, as discussed at https://github.com/vapor/docs/issues/850#issuecomment-1621691102.


0xTim (approved):
Thanks!

0xTim:
Removing the `translation-update` as we should update the other translations as well

alemohamad:
> Removing the `translation-update` as we should update the other translations as well

@0xTim having the tag `updates` will trigger the issue with new content to be translated?

0xTim:
> > Removing the `translation-update` as we should update the other translations as well
> 
> @0xTim having the tag `updates` will trigger the issue with new content to be translated?

Any PR that is closed without the `translation-update` or `no-update-needed` will trigger the workflow

Issue #1038, opened 2024-12-21
Add swift-testing compatibility docs


MahdiBM:
See https://github.com/vapor/vapor/pull/3257 and https://github.com/vapor/template/pull/147.

Mention `.testable()` -> `.testing()`.
Mention `XCTVapor` -> `VaporTesting`.
Mention you should explicitly `import Testing`.
Mention you MUST use the corresponding module or there is a good chance your Vapor test failures won't get reported.

alemohamad:
@MahdiBM with the new merged PR, can this be closed? Thanks!

MahdiBM:
Thank you.

Issue #594, opened 2021-03-18
Provide database integration tests best practices


0xTim:
We should add a section in the docs for best practices and options when running integration tests with a real database.

Points to touch on should include:

* setting up the database configuration for testing (you don't want to use the live DB!)
* using `autoRevert()`/`autoMigrate()` in tests
* using a transaction in tests to speed them up

alemohamad:
@0xTim  with the new merged PR, can this be closed? Thanks!

0xTim:
Yes lets close for now and can create separate issues if there's demand

Issue #865, opened 2023-07-23
Remove text about precondition failure on shutdown as this has been resolved.


BennyDeBock:
According to @Joannis, the text about having to call `app.shutdown` in tests is obsolete as this behaviour has been fixed.
Thus it should be removed

[Conversation](https://github.com/vapor/docs/pull/864)

This was added in #809

gwynne:
The warning was never rendered outdated, merely made inaccurate—it should say that failing to call shutdown will result in warnings. Calling it is still considered a requirement.

PR #1054, opened 2025-01-15
Update Testing documentation

alemohamad:
Following the issues #1038 and #594, here's a proposal for the update of Testing documentation, changing XCTest to Swift Testing.


MahdiBM (commented):
This looks good but not sure if we want to replace the XCT docs, or add a new one for Swift Testing.

0xTim (changes requested):
I agree with @MahdiBM - we should keep the old XCTVapor docs but put a warning telling people they should choose VaporTesting unless they have a reason not to

alemohamad:
I understand. In my opinion we should be promoting the Swift Testing option first and then add a note about the XCTest framework. I think that having a complete example with the new framework can be another way of encouraging its usage.

What do you prefer? Maintain the XCTest with a note about Swift Testing or the other way around?

0xTim:
> What do you prefer? Maintain the XCTest with a note about Swift Testing or the other way around?

Swift Testing first, XCTest at the bottom

docs/advanced/testing.md
```diff
@@ -1,99 +1,199 @@
...
+
+Then, add `import XCTVapor` at the top of your test files. Create classes extending `XCTestCase` to write tests cases.
+
+```swift
+import XCTVapor
+
+final class MyTests: XCTestCase {
+    func testStub() throws {
+        // Test here.
+    }
```
0xTim, line 194: Can we just copy over the old docs? So basically what we had, plus the note, about preferring VaporTesting (but add it at the top) with the new Swift Testing blurb at the top?

0xTim (changes requested):

docs/advanced/testing.md
alemohamad: Sure! Let me update the PR.

docs/advanced/testing.md
```diff
@@ -1,8 +1,176 @@
 # Testing
 
+## VaporTesting
+
+Vapor includes a module named `VaporTesting` that provides test helpers built on `Swift Testing`. These testing helpers allow you to send test requests to your Vapor application programmatically or running over an HTTP server.
+
+!!! note
+    For newer projects or teams adopting Swift concurrency, `VaporTesting` is highly recommended due to its simplicity and integration with Vapor.
```
MahdiBM, line 8: Not sure what exactly would be the best but I'd want to sneak in a mention of "XCT" in there just so it's clear what it's recommended over. Maybe something like this but not sure:
```suggestion
!!! note
    For newer projects or teams adopting Swift concurrency, `Swift Testing` is highly recommended over `XCT`.
```

alemohamad: I like the suggestion. I don't know what's better, to say Swift Testing or VaporTesting. And the other part I'll name it XCTest or XCTVapor, instead of XCT. What do you think? cc/ @0xTim

alemohamad:
@0xTim now that we are reviewing the Testing page, should we look at this issue #865 ? We can remove the text about precondition failure, if this is resolved. Let me know.

0xTim:
@alemohamad you still need to call it to shutdown services, but if you use the `withApp` stuff then it's done for you

docs/advanced/testing.md
0xTim: `XCT` doesn't really mean anything, so for this instance we should say `XCTest`

docs/advanced/testing.md
```diff
@@ -1,8 +1,176 @@
...
+
+To ensure your tests run in a serialized manner (e.g., when testing with a database), include the `.serialized` option in the test suite declaration:
+
+```swift
+@Suite("App Tests with DB", .serialized)
+```
+
+#### Running Tests
+
+Use `cmd+u` with the `-Package` scheme selected to run tests in Xcode. Use `swift test --enable-test-discovery` to test via the CLI.
```
0xTim, line 59: Test discovery is no longer a thing

docs/advanced/testing.md
```diff
@@ -1,8 +1,176 @@
...
+
+Each function marked with `@Test` will run automatically when your app is tested.
+
+To ensure your tests run in a serialized manner (e.g., when testing with a database), include the `.serialized` option in the test suite declaration:
+
+```swift
+@Suite("App Tests with DB", .serialized)
+```
+
+#### Running Tests
```
0xTim, line 57: I don't think we really need this, open to keeping it though. The doc isn't about how to run tests, it's how to use Vapor Testing

alemohamad: Yeah, I kept it because it was already there, but I'm happy to remove this.

docs/advanced/testing.md
alemohamad: Glad to hear this. I'll remove this.

0xTim (approved):
Thanks!

alemohamad:
@0xTim sorry, tapped on cancel the auto-merge.

PR #1060, opened 2025-02-04
QueuesFluentDriver is now an officially supported driver

petrpavlik:
- Update the docs to say that `QueuesFluentDriver` is not an officially supported driver alongside `QueuesRedisDriver`
- Link to the fork under `vapor-community` org instead of the original repo


0xTim:
It's still in the community org so not part of the core org so I think we should probably keep it under the community list. We should definitely update the link though

petrpavlik:
Thanks, makes sense. I'll open a new PR with just the link updated.

PR #1068, opened 2025-03-08
Reordered languages & added original names in translation_needed

alemohamad:
This proposal is to have the original languages in the `translation_needed.description.leaf` file, and also reorder the languages in both `translation_needed` and `mkdocs.yml` to follow the **ISO 639-1** language codes as used in `fixSearchIndex.swift` file, to have a unified order.


0xTim (approved):
Thanks!

PR #1073, opened 2025-03-25
Fix "Model Coupling" example in Schema docs

pkillian:
Amends the `UserNameMigration` code example in the "Model Coupling" section of the Schema documentation with correct syntax. Added closing square brackets to the dictionary literals passed into `QueryBuilder.set(_ data:)` in multiple places.


ptoffy (approved):
Thanks!

PR #1075, opened 2025-04-04
Suggest "Vapor for VS Code" extension

fpseverino:
- Replace "Leaf HTML" extension with "Vapor for VS Code"
- Add a new tip in the "Hello World" section suggesting the extension


0xTim (approved):
Thanks!

PR #1117, opened 2026-04-08
Update docs: updated readFile version for async await

swifty-amay:
###Changes

- Updated the depricated version of readFile method for reading files asynchronously.
- Changes 
try await req.fileio.readFile(at: "/path/to/file") { chunk in
        print(chunk) // ByteBuffer
  }
(Depricated version)

    to 

    let file = try await request.fileio.readFile(at: path, chunkSize: 16 * 1024) // 32Kb, ~5 chunks


ptoffy (changes requested):
Thanks for this! We can expand it to loop over the chunks AsyncSequence which will be what most users will do

swifty-amay:
Thanks for the review, actually you're right, devs must also get to know how to loop over the chunks.

ptoffy (changes requested):
`readContent` is undefined here. Let's just keep the behaviour we had before where we simply print the ByteBuffer chunk

swifty-amay:
Yes, you're right. Well, let me see if I can do something with this readContent, if not then I'll update this to the previous version where the chunks were being printed directly.

PR #1130, opened 2026-06-04
Migrate to Kiln

0xTim:
Migrate the docs site to Kiln. This is a tool I've whipped up (sorry Gwynne...) to take over MKdocs since that is in a weird place and gives us more control, and will allow us to customise the layout without it being behind a paywall.

Features:

* It's Swift, with compile time config and link checking
* Localisation support
* Actual strings localisation for the different strings in the site
* Better search that works across languages and supports accent searching etc
* Support for custom error pages
* Opengraph support etc

PR #1133, opened 2026-06-10
New Design

0xTim:
Now we control the doc generation stack we can (finally!) adopt the new design on the docs site.

Screenshot 2026-06-09 at 19 56 36


ptoffy (commented):
Nice! Adding them as I see them
- heading margin from the divider should be the same as the one above the divider:
    Image
- divider width should be consistent
    Image
- should we make boxes border radii consistent? (we definitely should for multi-line vs single-line code boxes)
    Image
- we might want some fade effects on text hover in the left sidebar and toc like we have in the navbar
- multiline codeblocks need padding resizing on top, bottom and left
- we probably want copy-code buttons
- what's going on here?
    Image

docs/4.0/index.md
TheHandyOwl, line 10: @0xTim, does this `</div>` go here? Or should it go before the H1 heading?

Do the rest of the documents need updating? This `<div>...</div>` is missing in the other translation files (index.XX.md)

[FIle updated here](https://github.com/vapor/docs/pull/1130/changes#diff-7d3080d3465150e733a45c8c9da93c7980265bc970843000fcd7ff85267657eb)

fpseverino:
Looking good!

Everything Paul said, but especially:
> we probably want copy-code buttons

Section, language and version selection is a bit weird on mobile IMHO

The Welcome page doesn't have the heading image and/or title in some languages, is it something we translators have to fix?

docs/4.0/index.md
0xTim: It's a slight hack but yes the div goes there, it pulls the title into that div. The others should have it will add

0xTim:
> Nice! Adding them as I see them

A lot of these are to be fixed in the design repo as that's where the shared styles live

0xTim:
> Everything Paul said, but especially:
> 
> > we probably want copy-code buttons

Yep planning to add this to the design repo so they all get it across all the sites

0xTim:
> Section, language and version selection is a bit weird on mobile IMHO

Anything specific @fpseverino ?

fpseverino:
> > Section, language and version selection is a bit weird on mobile IMHO
> 
> 
> 
> Anything specific @fpseverino ?

image

The button to open the section selection panel (circled in red) is a bit small and easy to miss (I have a big phone), and the language and version selection being in the same menu (circled in green) where the links to the home and blog are is a bit weird IMHO, it's not where I would intuitively think they are.

Take everything I say with a pinch of salt, I'm not a designer (pretty sure in fact I butchered the terminology)

0xTim:
> The button to open the section selection panel (circled in red) is a bit small and easy to miss

That's a fair point. I've bumped the size

As for the placement of the other options, the main issue with putting them in the left hand menu is that they'll be pushed down and not visible for most of the time. We can iterate further though!

PR #1136, opened 2026-06-16
Enable the real CSP

0xTim:


0xTim:
Going to leave this for now - we're gonna have issues with the Carbon ads as they load images from wherever

PR #1154, opened 2026-07-22
Update README

swifty-amay:
A separate PR for the README file.


swifty-amay:
This is in reference to the change requested in README stated as 
"You shouldn't need Python anymore, we've migrated to Kiln. But let's update this bit in a separate PR so if you could remove this section, that would be great (keep the typo though!)"

But using 'swift run'  is only generating and giving a static ./site folder, so if python3 is not there then who will serve the folder?

Kiln is only  being used for building the folder named ./site as stated in ref.  to line 37 in
' docs/Sources/VaporDocs/main'

                   ' try await Kiln.build(site, contentDirectory: "docs", outputDirectory: outputDirectory, linkChecking: .error,    
                    leafTags: VaporDesignTheme.leafTags) '

So what should I do now?

image

fpseverino:
> who will serve the folder?

I think we should tell people to use the [Kiln CLI](https://kiln.brokenhands.io/getting-started/cli/#kiln-serve), right @0xTim?

swifty-amay:
Ya you're right 
the 'swift run kiln serve' command is working fine
Let me just update it.

Thanks for sharing the info @fpseverino

README.md
```diff
@@ -18,26 +18,24 @@ http://127.0.0.1:8000
...
-3. then
-```
-cd site
-```
-4. then
+
+3. then run
 ```
-python3 -m http.server 8000
+swift run kiln serve
```
0xTim: This should be `kiln serve` not `swift run kiln serve`. We should also link to Kiln so users know how to install the CLI

README.md
```diff
@@ -18,26 +18,24 @@ http://127.0.0.1:8000
 ```
 
 # Start with SPM
-* Prerequisite for this setup, you must have swift and python installed on the system
+* Prerequisite for this setup, you must have swift installed on the system
 
 1. Clone the repo
 
-2. then run
+2. the move to the folder where you cloned the repo
```
0xTim: I think we can combine 1 and 2 and just say clone the repo and navigate to the cloned directory

0xTim (changes requested):

README.md
swifty-amay: Added this.

README.md
swifty-amay: Yes you're right, both can be combined in a single step.

0xTim (approved):
Thanks!

PR #1164, opened 2026-08-23
add documentation for index-variable

AntVil:
Document `#(index)` variable

### DIFF
--- a/docs/4.0/leaf/overview.md
+++ b/docs/4.0/leaf/overview.md
@@ -154,7 +154,7 @@ Planets:
 - Mars
 ```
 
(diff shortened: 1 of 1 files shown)


docs/4.0/leaf/overview.md
```diff
@@ -154,6 +154,26 @@ Planets:
 - Mars
 ```
 
+Inside the scope of the `#for` tag the implicit variable `index` is definded and can be accessed with `#(index)`. Naming the variable explicitly is also possible like this:
```
0xTim:
```suggestion
Leaf also offers an index-entry option for iteration which allows you to access the index in the loop:
```

0xTim (approved):
Thanks!

